PR #17, opened 2021-12-15
Fixed bug to get ipu_config.json from_pretrained path

jimypbr:
Before to get a pretrained ipu_config you had to specify the json file directly:

```
IPUConfig.from_pretrained("/path/to/config/ipu_config.json")
```

Model hub doesn't let you do this and it's not the behaviour of `from_pretrained` for other configs etc.

Now we can just do

```
IPUConfig.from_pretrained("/path/to/config")
```


michaelbenayoun:
Nice contribution, thanks!

michaelbenayoun (approved):

PR #31, opened 2022-02-04
Hook removal bugfix

jimypbr:
There's a bug when running BERT and RoBERTa fine-tuning models where after training is finished and you attempt to compile the inference model you get a tracing error in poptorch:

```
First diverging operator:
        Node diff:
                - %qa_outputs : __torch__.poptorch.ops.___torch_mangle_628.Linear = prim::GetAttr[name="qa_outputs"](%self.1)
                ?                                                      ^ -
                + %qa_outputs : __torch__.poptorch.ops.___torch_mangle_842.Linear = prim::GetAttr[name="qa_outputs"](%self.1)
                ?                                                      ^^
```

This was root-caused to us not removing the `outline_attributes` hooks in the `deparallelize` method for these models.
I have fixed this and also done some refactoring of RoBERTa and BERT models to remove repeated code.

### DIFF
--- a/optimum/graphcore/models/bert/modeling_bert.py
+++ b/optimum/graphcore/models/bert/modeling_bert.py
@@ -117,7 +117,7 @@ def recomputation_checkpoint(module: nn.Module):
     def recompute_outputs(module, inputs, outputs):
         return tuple(poptorch.recomputationCheckpoint(y) for y in outputs)
 
-    module.register_forward_hook(recompute_outputs)
+    return module.register_forward_hook(recompute_outputs)
 
 
 def outline_attribute(module: nn.Module, value: str):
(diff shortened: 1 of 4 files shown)


optimum/graphcore/models/bert/modeling_bert.py
```diff
@@ -358,21 +348,18 @@ def forward(self, indices):
...
         - (If enabled) Replaces the word embedding with a SerializedEmbedding
         - Adds recomputation checkpoints
-
-        Recommended usage:
-        ```
-        model = PipelinedBertForSequenceClassification(config).parallelize().half()
-        ```
         """
+        super().parallelize()
+        self._hooks = []
```
michaelbenayoun, line 361: Isn't it already existing?
If not I think we can actually put that in `PipelineMixin.parallelize`, wdyt?

michaelbenayoun (approved):
Great refactoring, thanks @jimypbr !

optimum/graphcore/models/bert/modeling_bert.py
jimypbr: Done.

Issue #35, opened 2022-02-17
Text classification example fails on several tasks


juliensimon:
Using the latest source install (d8b082ab341678ff2380eb15377a5ceea9c7c7f6), the first snippet in the text classification example fails on three tasks( mrpc, rte and wnli) with the same error message. 

transformers.__version__
'4.16.2'
datasets.__version__
'1.18.3'

Other tasks run fine on the IPU. The three impacted tasks also run fine with vanilla transformers.

```
python run_glue.py \
  --model_name_or_path bert-base-cased \
  --ipu_config_name Graphcore/bert-base-ipu \
  --task_name $TASK_NAME \
  --do_train \
  --do_eval \
  --max_seq_length 128 \
  --per_device_train_batch_size 32 \
  --learning_rate 2e-5 \
  --num_train_epochs 3 \
  --output_dir ./output/$TASK_NAME/

huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
To disable this warning, you can either:
	- Avoid using `tokenizers` before the fork if possible
	- Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true | false)
The following columns in the training set  don't have a corresponding argument in `PipelinedBertForSequenceClassification.forward` and have been ignored: sentence1, sentence2, idx.
[08:55:10.779] [poptorch::python] [critical] StopIteration

Traceback (most recent call last):
  File "run_glue.py", line 568, in <module>
    main()
  File "run_glue.py", line 486, in main
    train_result = trainer.train(resume_from_checkpoint=checkpoint)
  File "/home/juliens/optimum-env/lib/python3.6/site-packages/optimum/graphcore/trainer.py", line 986, in train
    self._compile_model(model, next(iter(train_dataloader)), log=True)
  File "/home/juliens/optimum-env/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 521, in __next__
    data = self._next_data()
  File "/home/juliens/optimum-env/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 560, in _next_data
    index = self._next_index()  # may raise StopIteration
  File "/home/juliens/optimum-env/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 512, in _next_index
    return next(self._sampler_iter)  # may raise StopIteration
StopIteration

02/17/2022 08:55:10 - critical - poptorch::python - StopIteration

Traceback (most recent call last):
[...]

PR #36, opened 2022-02-17
Update text classification example

juliensimon:
- Replace mrpc task (buggy right now, see https://github.com/huggingface/optimum-graphcore/issues/35) with sst2
- Fix name of IPU config file on the hub
- Delete local IPU config file
- Update requirements.txt

This was successfully tested with the latest source install (d8b082ab341678ff2380eb15377a5ceea9c7c7f6)


jimypbr (approved):
thanks

PR #42, opened 2022-02-18
Add masked language modeling

gejinchen:


jimypbr:
Please add instructions for running roberta mlm in the readme.

gejinchen:
> what's wrong with first way? Also just delete the code instead of commenting.

I suspected this could be the cause for the accum_type error. Now I know it's not.

optimum/graphcore/trainer.py
```diff
@@ -1201,7 +1201,8 @@ def train(
     def _load_state_dict_in_model(self, state_dict):
         self.model.deparallelize()
         load_result = self.model.load_state_dict(state_dict, strict=False)
-        self.model.parallelize()
+        if not self.args.fp32:
+            self.model.parallelize().half()
```
jimypbr, line 1205: This isn't quite right
Should be:
```
self.model.parallelize()
if not self.args.fp32:
    self.model.half()
```

optimum/graphcore/trainer.py
```diff
@@ -1203,23 +1203,21 @@ def train(
...
         # if self.inference_model and self.inference_model.isAttachedToDevice():
         #     self.inference_model.copyWeightsToDevice()
-
         if len(load_result.missing_keys) != 0:
-            if self.model._keys_to_ignore_on_save is not None and set(load_result.missing_keys) == set(
+            if self.model._keys_to_ignore_on_save is not None and set(load_result.missing_keys) != set(
                 self.model._keys_to_ignore_on_save
             ):
-                self.model.tie_weights()
-                self._wrap_model(self.model, self.is_in_train)
```
michaelbenayoun, line 1221: I get why this should be removed, but why change the condition and remove the else statement?

optimum/graphcore/models/roberta/modeling_roberta.py
```diff
@@ -80,6 +115,88 @@ def deparallelize(self):
         return self
 
 
+@register(RobertaForMaskedLM)
+class PipelinedRobertaForMaskedLM(RobertaForMaskedLM, PipelineMixin):
```
michaelbenayoun, line 119: That's great!
Maybe we can do some refactoring like we did for Bert, with some kind of `RobertaPipelineMixin`?

optimum/graphcore/models/roberta/modeling_roberta.py
```diff
@@ -37,6 +40,38 @@
 logger = logging.get_logger(__name__)
 
 
+class OnehotGather(nn.Module):
```
michaelbenayoun, line 43: Do you think we should add this to modeling_utils? I think this is also used for Bert.

optimum/graphcore/ipu_configuration.py
```diff
@@ -266,3 +267,44 @@ def batch_size_factor(self, for_inference: bool = False, pod_type: Optional[str]
         gradient_accumulation_steps = 1 if for_inference else ipu_config.gradient_accumulation_steps
         device_iterations = ipu_config.inference_device_iterations if for_inference else ipu_config.device_iterations
         return replication_factor * gradient_accumulation_steps * device_iterations
+
+    def update_from_string(self, update_str: str):
```
michaelbenayoun, line 271: Why is this needed?

examples/language-modeling/run_mlm.py
```diff
@@ -0,0 +1,565 @@
...
+    )
+    use_auth_token: bool = field(
+        default=False,
+        metadata={
+            "help": "Will use the token generated when running `transformers-cli login` (necessary to use this script "
+            "with private models)."
+        },
+    )
+
+    # def __post_init__(self):
```
michaelbenayoun, line 114: Should this be commented?

examples/language-modeling/README.md
```diff
@@ -47,11 +47,43 @@ python run_pretraining.py \
...
 When training a model from scratch, configuration values may be overridden with the help of `--config_overrides`:
 
 
 ```bash
 python run_pretraining.py --config_overrides="hidden_size=1024,num_attention_heads=16,num_hidden_layers=24" [...]
 ```
+
+## RoBERTa and masked language modeling
+
+The following example fine-tunes RoBERTa-base on WikiText-2. We're using the raw WikiText-2. Note the some IPU configurations are overridden.
```
michaelbenayoun, line 61: I think there is a typo "Note the some IPU..."

optimum/graphcore/ipu_configuration.py
jimypbr: I think that the ipu_config_overrides didn't work for parameters with List type.

gejinchen: Right, this is to support overriding the `matmul_proportion` config.

optimum/graphcore/models/roberta/modeling_roberta.py
gejinchen: There is already a `RobertaPipelineMixin`.

optimum/graphcore/trainer.py
jimypbr: I was just trying to invert the case fall back to the warning about missing keys. I could just stick the warning under the if statement on 1216, on second viewing, or?

michaelbenayoun: Ok, I get it!
And `self.model.tie_weights()` is not needed because we already do that in `parallelize` right?

michaelbenayoun: Ok, I get it!
And `self.model.tie_weights()` is not needed because we already do that in `parallelize` right?

jimypbr: Yes it is

optimum/graphcore/models/roberta/modeling_roberta.py
jimypbr: Yes. I can do that in this pr too.

examples/language-modeling/README.md
gejinchen: Yes you are right. Thanks.

optimum/graphcore/models/roberta/modeling_roberta.py
gejinchen: Yes! I will relocate it.

examples/language-modeling/run_mlm.py
jimypbr: Yeah it's commented out in `run_pretraining.py`

PR #53, opened 2022-02-22
Update text classification example

juliensimon:
- Add note on correct batch size for some tasks
- These values have been correctly tested

This is a fix for https://github.com/huggingface/optimum-graphcore/issues/35

### DIFF
--- a/examples/text-classification/README.md
+++ b/examples/text-classification/README.md
@@ -43,7 +43,7 @@ python run_glue.py \
   --output_dir ./output/$TASK_NAME/
 ```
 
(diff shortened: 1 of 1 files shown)


examples/text-classification/README.md
```diff
@@ -43,7 +43,7 @@ python run_glue.py \
   --output_dir ./output/$TASK_NAME/
 ```
 
-where task name can be one of cola, sst2, mrpc, stsb, qqp, mnli, qnli, rte, wnli.
+where task name can be one of cola, sst2, mrpc, stsb, qqp, mnli, qnli, rte, wnli. All tasks should work correctly with a batch size of 32, expect mrpc and rte (16) and wnli (4).
```
jimypbr, line 46: "except" rather than "expect"

PR #56, opened 2022-02-22
Add DeBERTa

jimypbr:
Waiting on next poplar SDK to land to main
- [x] Get base to compile and fit on IPU
- [x] Get base to train on SQuAD

### DIFF
--- a/optimum/graphcore/trainer.py
+++ b/optimum/graphcore/trainer.py
@@ -866,7 +866,16 @@ class IPUTrainer:
             trial = None
         self.state.is_hyper_param_search = trial is not None
 
+        # Activate gradient checkpointing if needed
+        if args.gradient_checkpointing:
+            self.model.gradient_checkpointing_enable()
+
         model = self._wrap_model(self.model)
+        self.model_wrapped = model
+
+        # TODO: handle optimizer and scheduler creation
+        # if delay_optimizer_creation:
+        #     self.create_optimizer_and_scheduler(num_training_steps=max_steps)
 
         # Check if saved optimizer or scheduler states exist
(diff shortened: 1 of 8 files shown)


optimum/graphcore/trainer.py
```diff
@@ -1208,14 +1204,11 @@ def _load_state_dict_in_model(self, state_dict):
...
 
-        # TODO: check if this is needed.
-        # if self.training_model and self.training_model.isAttachedToDevice():
-        #     self.training_model.copyWeightsToDevice()
-
-        # if self.inference_model and self.inference_model.isAttachedToDevice():
-        #     self.inference_model.copyWeightsToDevice()
         if len(load_result.missing_keys) != 0:
-            logger.warn(f"There were missing keys in the checkpoint model loaded: {load_result.missing_keys}.")
+            if self.model._keys_to_ignore_on_save is not None and set(load_result.missing_keys) != set(
```
michaelbenayoun, line 1208: @jimypbr I saw that you removed those tests when I rebased, should I keep those? 
What was the rational when removing them?

jimypbr: I think this is right to put them back. The check got removed when I was fixing the roberta checkpointing bug where there was an extra `tie_weights` that was breaking things.

michaelbenayoun (commented):
Why?
I think the changes we've done were good no?
The idea is to simplify things.

jimypbr:
> Why? I think the changes we've done were good no? The idea is to simplify things.

These changes were breaking the checkpointing with BERT. It was causing an error after it dumps the first checkpoint.

michaelbenayoun:
Ok, we will work on that in later PRs then.

PR #59, opened 2022-02-25
[WIP] Add LXMERT for visual question answering

gejinchen:
1. Still need to refactor the datasets and change the code accordingly.
2. Convergence runs on whole datasets are yet to be done.

### DIFF
--- a/examples/question-answering/README.md
+++ b/examples/question-answering/README.md
@@ -56,3 +56,22 @@ f1 = 88.52
 exact_match = 81.22
 ```
 -->
+
+# Visual question answering
+
+`run_vqa.py` allows you to fine-tune the LXMERT model for visual question answering datasets such as VQA v2 and GQA, which are available on our [hub](https://huggingface.co/datasets).
+
+## Fine-tuning LXMERT on GQA
+
+```bash
+python run_vqa.py \
+  --model_name_or_path unc-nlp/lxmert-base-uncased \
+  --ipu_config_name ipu_config_lxmert \
+  --dataset_name Graphcore/gqa-lxmert \
+  --do_train \
+  --do_eval \
+  --per_device_train_batch_size 1 \
+  --num_train_epochs 4 \
+  --learning_rate 5e-5 \
+  --output_dir /tmp/lxmert_gqa/ \
+```
--- a/examples/question-answering/README.md
+++ b/examples/question-answering/README.md
@@ -64,14 +64,23 @@ exact_match = 81.22
 ## Fine-tuning LXMERT on GQA
 
 ```bash
-python run_vqa.py \
+python examples/question-answering/run_vqa.py \
   --model_name_or_path unc-nlp/lxmert-base-uncased \
-  --ipu_config_name ipu_config_lxmert \
+  --ipu_config_name Graphcore/lxmert-base-ipu \
   --dataset_name Graphcore/gqa-lxmert \
   --do_train \
   --do_eval \
+  --max_seq_length 128 \
   --per_device_train_batch_size 1 \
   --num_train_epochs 4 \
-  --learning_rate 5e-5 \
-  --output_dir /tmp/lxmert_gqa/ \
+  --dataloader_num_workers 64 \
+  --logging_steps 5 \
+  --learning_rate 1e-5 \
+  --lr_scheduler_type linear \
+  --loss_scaling 16384 \
+  --weight_decay 0.01 \
+  --warmup_ratio 0.1 \
+  --output_dir /tmp/gqa/ \
+  --dataloader_drop_last \
+  --replace_qa_head
 ```
--- a/examples/question-answering/README.md
+++ b/examples/question-answering/README.md
@@ -70,7 +70,7 @@ python examples/question-answering/run_vqa.py \
   --dataset_name Graphcore/gqa-lxmert \
   --do_train \
   --do_eval \
-  --max_seq_length 128 \
+  --max_seq_length 512 \
   --per_device_train_batch_size 1 \
   --num_train_epochs 4 \
   --dataloader_num_workers 64 \
@@ -82,5 +82,6 @@ python examples/question-answering/run_vqa.py \
   --warmup_ratio 0.1 \
   --output_dir /tmp/gqa/ \
   --dataloader_drop_last \
-  --replace_qa_head
+  --replace_qa_head \
+  --pod_type pod16
 ```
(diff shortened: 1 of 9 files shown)


examples/question-answering/README.md
```diff
@@ -56,3 +56,31 @@ f1 = 88.52
...
 -->
+
+# Visual question answering
+
+`run_vqa.py` allows you to fine-tune the LXMERT model for visual question answering datasets such as VQA v2 and GQA, which are available on our [hub](https://huggingface.co/datasets).
+
+## Fine-tuning LXMERT on GQA
+
+```bash
+python examples/question-answering/run_vqa.py \
```
michaelbenayoun, line 67: Please add the pod type.

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
...
+    HfArgumentParser,
+    default_data_collator,
+    set_seed,
+)
+from transformers.trainer_utils import get_last_checkpoint
+from transformers.utils import check_min_version
+from transformers.utils.versions import require_version
+
+
+# from utils_qa import postprocess_qa_predictions
```
michaelbenayoun, line 53: Why is this commented?

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
...
+import datasets
+import numpy as np
+from datasets import load_dataset, load_metric
+
+import transformers
+from optimum.graphcore import IPUConfig, IPUTrainer
+from optimum.graphcore import IPUTrainingArguments as TrainingArguments
+from optimum.graphcore.data import pad_on_batch_axis
+
+# from trainer_qa import QuestionAnsweringTrainer
```
michaelbenayoun, line 37: Are you sure you dont want to use the `QuestionAnsweringTrainer`?

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
+#!/usr/bin/env python
```
michaelbenayoun, line 1: You will need to add the reference diff file to get tests to pass, I can help you with that.

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
...
+    )
+
+    log_level = training_args.get_process_log_level()
+    logger.setLevel(log_level)
+    datasets.utils.logging.set_verbosity(log_level)
+    transformers.utils.logging.set_verbosity(log_level)
+    transformers.utils.logging.enable_default_handler()
+    transformers.utils.logging.enable_explicit_format()
+
+    # Log on each process the small summary:
```
michaelbenayoun, line 241: You can remove those comments.

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
...
+    # Data collator
+    # We have already padded to max length if the corresponding flag is True, otherwise we need to pad in the data
+    # collator.
+    data_collator = (
+        default_data_collator
+        if data_args.pad_to_max_length
+        else DataCollatorWithPadding(tokenizer, pad_to_multiple_of=8 if training_args.fp16 else None)
+    )
+
+    # if training_args.do_train and training_args.pad_on_batch_axis:
```
michaelbenayoun, line 425: Why is this commented? Should we remove this?

optimum/graphcore/models/lxmert/modeling_lxmert.py
```diff
@@ -0,0 +1,113 @@
...
+        logger.info("Image embedding         --> IPU 0")
+        self.lxmert.encoder.visn_fc = poptorch.BeginBlock(self.lxmert.encoder.visn_fc, "Image embedding", ipu_id=0)
+
+        # Language layers
+        for index, layer in enumerate(self.lxmert.encoder.layer):
+            if self.config.recompute_checkpoint_every_layer:
+                h = recomputation_checkpoint(layer)
+                self._hooks.append(h)
+            self.lxmert.encoder.layer[index] = poptorch.BeginBlock(layer, f"Language layer{index}", ipu_id=1)
+            logger.info(f"Language layer {index:<2}       --> IPU 1")
```
michaelbenayoun, line 55: Why are ipu_ids hardcoded?

optimum/graphcore/models/lxmert/modeling_lxmert.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        input_ids,
+        visual_feats,
+        visual_pos,
+        attention_mask=None,
+        token_type_ids=None,
+        labels=None,
+        visual_attention_mask=None,
+    ):
+        lxmert_output = self.lxmert(
```
michaelbenayoun, line 91: Are you changing the forward pass implementation or can we use the `super()`  method?

gejinchen: Yes, the forward() is changed. The only difference is that now the forward() could support soft labels from the VQA v2 dataset.

examples/question-answering/run_vqa.py
```diff
@@ -0,0 +1,532 @@
...
+
+    if training_args.push_to_hub:
+        trainer.push_to_hub(**kwargs)
+    else:
+        trainer.create_model_card(**kwargs)
+
+
+def _mp_fn(index):
+    # For xla_spawn (TPUs)
+    main()
```
jimypbr, line 528: remove this

examples/question-answering/run_vqa.py
gejinchen: I am pretty sure. VQA and SQuAD are actually quite different tasks. VQA's prediction head just outputs a single prediction, so it's more like image classification.

examples/question-answering/run_vqa.py
gejinchen: It should be deleted. VQA's outputs are not substrings of contexts.

examples/question-answering/run_vqa.py
gejinchen: It turned out to be quite complicated to handle the dataloader remainder for soft labels. Considering that most tasks just do drop_last, I think we should just do the same for VQA for now.

optimum/graphcore/models/lxmert/modeling_lxmert.py
gejinchen: There is only LXMERT-base. No small or medium ... One could of course override and add more layers to each component, and since the components are not the same in memory consumption, one of the IPUs might go OOM earlier than others. Therefore, this arrangement may not be the most memory efficient.

michaelbenayoun (approved):

PR #61, opened 2022-03-03
Model ipu config attribute

michaelbenayoun:
# What does this PR do?

Currently, the `IPUConfig` dict is merged into the `model.config` attribute. This is bad because when saving the model and its configuration, the IPU config content is saved in the model configuration at the same time. 

To avoid that, an `ipu_config` attribute needs to be added to pipelined models. It's already be taken care of with the class methods from `PipelineMixin`, and for the case in which a pipeline object is instantiated from the class constructor, the user should get an error when trying to parallelize the model if the attribute was not set beforehand.

We could also add an optional `ipu_config` argument to the parallelize method to allow the user to provide this config directly instead of doing:
```python
pipelined_model.ipu_config = IPUConfig.from_pretrained(....)
pipelined_model.parallelize()
``` 

@jimypbr wdyt?

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -86,9 +86,9 @@ class PipelineMixin:
     @classmethod
     def from_model(cls, model: nn.Module):
         clone = copy.deepcopy(model)
-        # It is fine because PipelineMixin only adds functionality, it does not add any attribute.
-        # (The only attribute it adds is not used here)
         clone.__class__ = cls
+        # Just needed so that .parallelize() does not throw an error
+        clone.ipu_config = IPUConfig()
         return clone
 
     @property
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -93,16 +93,16 @@ class PipelineMixin:
 
     @property
     def ipu_config(self):
-        ipu_config_ = getattr(self, "ipu_config", None)
-        if ipu_config_ is None:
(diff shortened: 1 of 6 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -70,30 +70,44 @@ def to_pipelined(model: nn.Module, ipu_config: IPUConfig, force: bool = False):
...
+    @ipu_config.setter
+    def ipu_config(self, value: IPUConfig):
+        if not isinstance(value, IPUConfig):
+            raise TypeError(f"ipu_config must be an instance of IPUConfig, but {type(value)} was provided")
+        self._ipu_config = value
+
     def parallelize(self):
         """Transform the model to run in an IPU pipeline."""
         self._hooks = []
+        _ = self.ipu_config  # Just test that the ipu_config was set before calling parallelize
```
jimypbr, line 110: was going to suggest assert, but I see that this will raise an exception through the property. nice.

jimypbr (approved):
Yes this is cleaner to explictly separate the configs!

optimum/graphcore/modeling_utils.py
michaelbenayoun: Yes, maybe that's not clear enough?

michaelbenayoun: Just merged the PR and didn't see your comment, will integrate that now, sorry about this.

michaelbenayoun: A good one!

PR #62, opened 2022-03-04
Update generation model evaluation

regisss:
# What does this PR do?

Adds evaluation support for `BartForConditionalGeneration`.


regisss:
Continued in #76.

PR #65, opened 2022-03-10
Test and tool to check that transformers and optimum examples match

michaelbenayoun:
# What does this PR do?

It provides a test checking that the optimum version of the examples is matching the transformers one.
To do that the following is done for each example script:

1. We have a "diff" file located under `tests/examples/[name of the example script].txt`, this file acts as a reference of the differences the two files should have (Trainer vs IPUTrainer, IPUConfig stuff and so on)
2. A copy of the transformers repo is downloaded and from it the "today" difference between the transformers version and the version in optimum is computed.
3. We check that this up-to-date difference matches the reference diff file.

To create the reference diff file, a tool is provided under `tests/create_diff_file_for_example.py`:
```
python tests/create_diff_file_for_example.py \
 --transformers ~/transformers/examples/pytorch/text-classification/run_glue.py \
 --optimum examples/text-classification/run_glue.py
```

This will prompt the differences between the two files, and ask which one is "legit" interactively.

### DIFF
--- a/tests/create_diff_file_for_example.py
+++ b/tests/create_diff_file_for_example.py
@@ -56,6 +56,8 @@ def _colorize_lines(content):
     }
     end_color = "\033[0;0m"
     for i, line in enumerate(lines):
+        if not line:
+            continue
         start_char = color_mapping.get(line[0], color_mapping["default"])
         lines[i] = "".join([start_char, line, end_color])
     return "\n".join(lines)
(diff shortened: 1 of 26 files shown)


tests/create_diff_file_for_example.py
```diff
@@ -0,0 +1,115 @@
...
+import re
+import subprocess
+from argparse import ArgumentParser
+from pathlib import Path
+
+
+DIFF_DIRECTORY = Path("tests/examples")
+
+
+def _ask_yer_or_no_question(message: str) -> str:
```
jimypbr, line 26: yes*

jimypbr (approved):
Looks great. super useful tool 🥇

Issue #67, opened 2022-03-15
Support all the poptorch optimizers


jimypbr:
We currently only support AdamW and LAMB.

In IPUTrainer:

```
    def _pytorch_optimizer_to_poptorch(self, optimizer: optim.Optimizer):
        # TODO: implement this function
        pytorch_to_poptorch_mapping = {
            optim.SGD: poptorch.optim.SGD,
            optim.Adam: poptorch.optim.Adam,
            optim.AdamW: poptorch.optim.AdamW,
            optim.RMSpop: poptorch.optim.RMSprop,
        }
        pass
```

PR #68, opened 2022-03-17
Trainer cleanup

michaelbenayoun:
# What does this PR do?

- Clean and remove unsused code in the Trainer
- The `IPUTrainer` can now accept optimizers in the init and those will be converted to `poptorch.DataLoader`
- "New" tests
- Multiple bugs fixed

Fixes: #67

### DIFF
--- a/optimum/graphcore/trainer.py
+++ b/optimum/graphcore/trainer.py
@@ -960,10 +960,6 @@ class IPUTrainer:
 
             for step, inputs in enumerate(epoch_iterator):
 
-                # TODO: remove that.
-                # if epoch == 0 and step == 5:
-                #     import pdb; pdb.set_trace()
-
                 # Skip past any already trained steps if resuming training
                 if steps_trained_in_current_epoch > 0:
                     steps_trained_in_current_epoch -= 1
@@ -1260,11 +1256,8 @@ class IPUTrainer:
             :obj:`torch.Tensor`: The tensor with training loss on this batch.
         """
         inputs = self._prepare_inputs(inputs)
-        # logger.info(f"{model.a} {model.b}")
         loss = self.compute_loss(model, inputs)
         loss = loss.mean()
-        logger.info(f"{self.lr_scheduler.state_dict()}")
-        # logger.info(f"{inputs} ----- {model.a} {model.b} ----- {loss}")
         return loss
 
     def compute_loss(self, model, inputs, return_outputs=False):
--- a/optimum/graphcore/trainer.py
+++ b/optimum/graphcore/trainer.py
@@ -266,7 +266,12 @@ class IPUTrainer:
         # very last
         self._memory_tracker.stop_and_update_metrics()
 
-    def _pytorch_optimizer_to_poptorch(self, optimizer: optim.Optimizer, model: Union[PreTrainedModel, nn.Module], pipelined_model: Union[PreTrainedModel, nn.Module]) -> poptorch.optim.Optimizer:
+    def _pytorch_optimizer_to_poptorch(
+        self,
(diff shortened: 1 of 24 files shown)


michaelbenayoun:
Yes, you're right, will do!

callumm-graphcore:
_(Hi, James suggested I post these question as a comment on this PR)_

I wanted to flag up something that we hit while porting Wav2Vec2. From `optimum-graphcore/optimum/graphcore/training_args.py`, we have the following argument for training:
```
logging_nan_inf_filter: str = field(default=True, metadata={"help": "Filter nan and inf losses for logging."})
```
There were two problems we had with this:
1. The default is `True`, which meant that we missed that the loss of our model was NaN the first time we ran it.
2. The type of the argument is `str`. This meant that when I tried to pass `--logging_nan_inf_filter=False`, this was interpreted as the string `"False"`. As a non-empty string, this was interpreted as a boolean `True`, and so we didn’t see the NaN loss until we passed `--logging_nan_inf_filter=""`.

With this in mind, I have some questions:
1. Why is the default `True`? Is there a good reason to hide NaN/Inf from the user?
2. Could we at least change the type of the argument to `bool`?

optimum/graphcore/trainer.py
```diff
@@ -959,35 +853,21 @@ def train(
...
 
         if DebugOption.UNDERFLOW_OVERFLOW in self.args.debug:
-            # TODO: test if this works.
             debug_overflow = DebugUnderflowOverflow(self.model)  # noqa
 
         self.create_optimizer_and_scheduler(num_training_steps=max_steps)
 
         self.state = TrainerState()
+        if trial is not None:
+            raise ValueError("Hyperparameyer tuning is not supported by the IPUTrainer.")
```
jimypbr, line 862: spelling hyperparameter

optimum/graphcore/trainer.py
```diff
@@ -303,22 +266,78 @@ def __init__(
...
+            The converted poptorch optimizer.
+        """
+        first_order_type = torch.float16 if self.ipu_config.enable_half_first_order_momentum else torch.float32
+        optimizer_kwargs = {
+            "loss_scaling": self.args.loss_scaling,
+            "accum_type": first_order_type,
+            "first_order_momentum_accum_type": first_order_type,
+            "second_order_momentum_accum_type": torch.float32,
+        }
+        # TODO: disabled max_grad_norm because it make things fail, fix it.
```
jimypbr, line 295: what's the issue with max_grad_norm?

optimum/graphcore/trainer.py
```diff
@@ -303,22 +266,78 @@ def __init__(
...
+        max_grad_norm = self.args.max_grad_norm
+        self.args.max_grad_norm = None
         pytorch_to_poptorch_mapping = {
-            optim.SGD: poptorch.optim.SGD,
-            optim.Adam: poptorch.optim.Adam,
-            optim.AdamW: poptorch.optim.AdamW,
-            optim.RMSpop: poptorch.optim.RMSprop,
+            optim.SGD: (poptorch.optim.SGD, {"loss_scaling": self.args.loss_scaling}),
```
jimypbr, line 299: does loss scaling get passed to the other optimizers as well?

optimum/graphcore/trainer.py
michaelbenayoun: Thanks for catching this!

optimum/graphcore/trainer.py
michaelbenayoun: It raises an error, I think we need to fix that.

PR #75, opened 2022-03-29
Examples sync

### LINKED ISSUE #71: Fix transformers version to match with Graphcore's SDK
The Graphcore's SDK is tested for certain versions of `transformers` (at the moment the maximum one is 4.17.0). However, optimum requires the latest version  of `transformers` or otherwise throws an error — see for instance [this line in `examples/question-answering/run_qa.py`](https://github.com/huggingface/optimum-graphcore/blob/main/examples/question-answering/run_qa.py#L53), where it checks for the most recent version:
```
# Will error if the minimal version of Transformers is not installed. Remove at your own risks.
check_min_version("4.18.0.dev0")
```
Is there a way of setting a certain version of transformers in optimum? Ideally we'd like to be able to checkout a certain `transformers` version and not master. Thanks.

michaelbenayoun:
# What does this PR do?

This synchronizes the examples with their transformers counterparts, and adds the transformers package as a requirement to install from sources for each example.

The current issue is that optimum-graphcore requires the stable transformers package, but the examples often require installing from sources because they can us the latest changes. This was causing issues (#71), as users can end up with an "old" version of transformers when installing optimum-graphcore, making the example scripts fail.

There are multiple ways of solving this:

1. Stop checking for the latest transformers version in the examples. This can work, but since we try to match and synchronize the examples between the two libraries, it can be bug-prone.
2. Synchronizing the examples with the latest transformers release instead of the repo itself. This could work, but this is not the way they do in transformers, and it would force us to track and update the tests to make sure that they are in sync with the latest stable version.
3. Adding transformers installed from sources in the examples requirements. This is the approach taken in this PR, and it seems to be the one to make the most sense since it both follows the way it is done in transformers (the tests are guaranteed to work with transformers installed from sources), and puts the constraints on the version of transformers only in the examples, which is fair since this is the only part actually needing that.

@jimypbr @whobbes @gejinchen What do you think?

Fixes #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ spell run \
   --docker_image "graphcore/pytorch:latest" \
   --apt python3 \
   --apt git \
-  "pip install git+https://github.com/huggingface/transformers.git . && \
+  "pip install . && \
    pip install -r [PATH TO POTENTIAL REQUIREMENTS] && \
    [INSERT YOUR COMMAND BETWEEN THE QUOTES]"
 ```
@@ -128,7 +128,7 @@ spell run \
   --docker_image "graphcore/pytorch:latest" \
   --apt python3 \
   --apt git \
-  "pip install git+https://github.com/huggingface/transformers.git . && \
+  "pip install . && \
    pip install -r examples/text-classification/requirements.txt && \
    python3 examples/text-classification/run_glue.py \
    --model_name_or_path bert-base-cased \
--- a/examples/image-classification/requirements.txt
+++ b/examples/image-classification/requirements.txt
@@ -1,6 +1,5 @@
+git+https://github.com/huggingface/transformers.git # Requires latest version of transformers
 --find-links https://download.pytorch.org/whl/torch_stable.html
-transformers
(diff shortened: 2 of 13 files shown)


jimypbr:
Wouldn't it be we should be compatible with the latest transformers on pipy then we have a stable thing to compare against? Would the CI track the changes to examples on the latest master or the latest release tag of transformers?

michaelbenayoun:
I was more thinking to match exactly what is being done in transformers, which assume examples work for the latest master branch. 

We can try to track the the last released version too, this would mean:

1. Stop checking for min versions in the examples (or change those lines with the latest release)
2. Removing transformers from source from the examples requirements

I feel like both are kind of equivalent no?

jimypbr:
Yeah it is equivalent and this would solve #71. Is it good to have this transformers version floating like that? Do you think it could cause a situation where transformers changes and breaks our examples from one day to the next?

jimypbr:
Ok then. 👍 Let's merge this then.

PR #79, opened 2022-04-07
Disable replicated tensor sharding only when num replicas is 1

jimypbr:


michaelbenayoun:
Is it ready to be merged?

jimypbr:
Yes. Do you think this is the best way to do this? I disabled it in the actual poptorch option before, but it was breaking one of the unit tests.

michaelbenayoun:
I think it would be better in the `_to_option` method. 
In any case, I think we shoud log something to specifying that this value is being overriden, wdyt?

PR #87, opened 2022-04-25
Add check_min_version to examples

michaelbenayoun:
# What does this PR do?

This adds a version check to all the examples. This should prevent and notice users trying out the examples from a pypi installation that they should install from sources.


jimypbr (approved):
LGTM

PR #90, opened 2022-04-28
Improve generation

gejinchen:
# What does this PR do?

Some changes to improve text generation so that it would be easier to integrate the run_generation.py task.

## Before submitting

### DIFF
--- a/optimum/graphcore/generation_utils.py
+++ b/optimum/graphcore/generation_utils.py
@@ -1413,7 +1413,11 @@ class IPUGenerationMixin(GenerationMixin):
             if not self.config.is_encoder_decoder:
                 model_kwargs["attention_mask"] = model_kwargs["attention_mask"][:, :cur_len]
 
-            next_token_logits = outputs.logits[:, cur_len - 1, :]
+            if outputs.logits.dim() == 3:
+                next_token_logits = outputs.logits[:, cur_len - 1, :]
+            # If the dimension of logits is 2, then only the logits of the last non-padding token is returned, so no need to slice.
+            else:
+                next_token_logits = outputs.logits
 
             # Store scores, attentions and hidden_states when required
             if return_dict_in_generate:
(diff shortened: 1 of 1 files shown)


optimum/graphcore/generation_utils.py
```diff
@@ -1387,11 +1385,17 @@ def greedy_search(
...
         unfinished_sequences = input_ids.new(input_ids.shape[0]).fill_(1)
-        input_ids = self._pad_input_ids_to_max_length(input_ids, max_length)
-
-        cur_len = torch.argmax((input_ids == self.config.pad_token_id).to(torch.long), dim=1)[0]
+        cur_len = input_ids.shape[-1]
 
         while True:
+
+            input_ids = self._pad_tensors_to_max_len(input_ids, max_length, pad_token_id)
+            # For a seq2seq model such as BART, the "attention_mask" is the enocder/cross attention mask and it does not require padding.
```
michaelbenayoun, line 1393: Typoe enocder -> encoder

optimum/graphcore/generation_utils.py
```diff
@@ -1404,7 +1408,16 @@ def greedy_search(
...
             outputs = self._poptorch_outputs_to_model_outputs(outputs)
 
-            next_token_logits = outputs["logits"][:, cur_len - 1, :].to(torch.float32)
+            # Remove padding and restore to actual length
+            input_ids = input_ids[:, :cur_len]
+            if not self.config.is_encoder_decoder:
+                model_kwargs["attention_mask"] = model_kwargs["attention_mask"][:, :cur_len]
+
+            if outputs.logits.dim() == 3:
+                next_token_logits = outputs.logits[:, cur_len - 1, :]
```
michaelbenayoun, line 1417: Should it still be `cur_len - 1`?

gejinchen: For BART and T5, I didn't use gather to pick out the logits at the last position because it is actually more complicated than GPT2, and also because they only need 128 length anyway. That is why I use this if/else statement to differentiate the gather/no gather situations. Dimension is 3 means no gather, so the index should still be `cur_len - 1`.

optimum/graphcore/generation_utils.py
gejinchen: Thanks!

PR #94, opened 2022-05-12
Use seq2seq data collator always in the translation example

jimypbr:
* Fixes a missing arg error in the `run_translation.py` example


michaelbenayoun (approved):
Great, thanks @jimypbr !
We might need to do the same thing for summarization.

PR #97, opened 2022-05-19
Reload rng_state when checkpointing

gejinchen:
While saving the checkpoint, get the random number generator state before `deparallelize()` and reload it after `parallelize()`, so that checkpointing won't alter the random state and we could have reproducibility no matter what the frequency of checkpointing is.

## Before submitting


jimypbr:
Could you add a little unit test for this please so we don't regress later? I think we can do it in `test_trainer`. Don't need to do it for every model. Just the test model that's used in `test_trainer`.

jimypbr:
Hmm I think that unit testing this well would mean running the test for every model to be meaningful. Let's just land this and think about adding that later.

jimypbr (approved):

PR #100, opened 2022-05-30
Error message when trying to use generate

michaelbenayoun:
# What does this PR do?

This PR adds an error message when the user wants to use `--predict_with_generate` since it is not well supported for now.


jimypbr:
Turn off the generate tests

jimypbr (approved):
lgtm

PR #118, opened 2022-07-06
Use last_ipu variable instead of a fixed ipu id for ViT

gejinchen:
## Before submitting


jimypbr (approved):
good thanks

PR #120, opened 2022-07-08
eval data collator

joao-alex-cunha:
This PR allows users to specify a data collator that is applied only to the evaluation dataset. This allows the application of data augmentation only to training or evaluation datasets

### DIFF
--- a/optimum/graphcore/trainer.py
+++ b/optimum/graphcore/trainer.py
@@ -139,6 +139,7 @@ class IPUTrainer:
         ipu_config: IPUConfig = None,
         args: IPUTrainingArguments = None,
         data_collator: Optional[DataCollator] = None,
+        eval_data_collator: Optional[DataCollator] = None,
         train_dataset: Optional[Dataset] = None,
         eval_dataset: Optional[Dataset] = None,
         tokenizer: Optional[PreTrainedTokenizerBase] = None,
@@ -188,6 +189,7 @@ class IPUTrainer:
 
         default_collator = default_data_collator if tokenizer is None else DataCollatorWithPadding(tokenizer)
         self.data_collator = data_collator if data_collator is not None else default_collator
+        self.eval_data_collator = eval_data_collator if eval_data_collator is not None else self.data_collator
         self.train_dataset = train_dataset
         self.eval_dataset = eval_dataset
         self.tokenizer = tokenizer
@@ -243,6 +245,9 @@ class IPUTrainer:
 
         if not callable(self.data_collator) and callable(getattr(self.data_collator, "collate_batch", None)):
             raise ValueError("The `data_collator` should be a simple callable (function, class with `__call__`).")
+
+        if not callable(self.eval_data_collator) and callable(getattr(self.eval_data_collator, "collate_batch", None)):
+            raise ValueError("The `eval_data_collator` should be a simple callable (function, class with `__call__`).")
 
         if args.max_steps > 0:
             logger.info("max_steps is given, it will override any value given in num_train_epochs")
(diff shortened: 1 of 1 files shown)


blazejba (approved):

optimum/graphcore/trainer.py
```diff
@@ -568,7 +573,7 @@ def get_eval_dataloader(self, eval_dataset: Optional[Dataset] = None) -> poptorc
 
         # self.data_collator can be a data collator that was wrapped by pad_on_batch_axis, retrieving the original
         # data collator as the wrapped version is only wanted for training.
-        data_collator = getattr(self.data_collator, "__wrapped__", self.data_collator)
+        data_collator = getattr(self.eval_data_collator, "__wrapped__", self.data_collator)
```
michaelbenayoun, line 576: Not sure, this is intended, so double checking?
```suggestion
        data_collator = getattr(self.eval_data_collator, "__wrapped__", self.eval_data_collator)
```

michaelbenayoun (commented):
I think we could add this. 
What happens if no `eval_data_collator` is provided? Do we use `data_collator` or nothing at all?

optimum/graphcore/trainer.py
```diff
@@ -188,6 +189,7 @@ def __init__(
 
         default_collator = default_data_collator if tokenizer is None else DataCollatorWithPadding(tokenizer)
         self.data_collator = data_collator if data_collator is not None else default_collator
+        self.eval_data_collator = eval_data_collator if eval_data_collator is not None else self.data_collator
```
joao-alex-cunha, line 192: When no eval_data_collator is provided by the user, then it is assumed that data_collator is used for both training and evaluation.

optimum/graphcore/trainer.py
joao-alex-cunha: Hi James, not sure I know enough Pytorch to understand what you are talking about.
You mean you want to allow the user to specify a different batch size for training and eval? How would this affect this part of the code?

jimypbr: self.data_collator can be a data collator that was wrapped by pad_on_batch_axis.
If no eval_data_collator is specified then use the unwrapped training data collator.

I suggest this change:

```
diff --git a/optimum/graphcore/trainer.py b/optimum/graphcore/trainer.py
index aeaf3b7..5297c1d 100644
--- a/optimum/graphcore/trainer.py
+++ b/optimum/graphcore/trainer.py
@@ -189,7 +189,9 @@ class IPUTrainer:
 
         default_collator = default_data_collator if tokenizer is None else DataCollatorWithPadding(tokenizer)
         self.data_collator = data_collator if data_collator is not None else default_collator
-        self.eval_data_collator = eval_data_collator if eval_data_collator is not None else self.data_collator
+        # self.data_collator can be a data collator that was wrapped by pad_on_batch_axis.
+        # If no eval_data_collator is specified then use the unwrapped training data collator.
+        self.eval_data_collator = eval_data_collator if eval_data_collator is not None else getattr(self.data_collator, "__wrapped__", self.data_collator)
         self.train_dataset = train_dataset
         self.eval_dataset = eval_dataset
         self.tokenizer = tokenizer
@@ -571,9 +573,7 @@ class IPUTrainer:
         if is_datasets_available() and isinstance(eval_dataset, datasets.Dataset):
             eval_dataset = self._remove_unused_columns(eval_dataset, description="evaluation")
 
-        # self.data_collator can be a data collator that was wrapped by pad_on_batch_axis, retrieving the original
-        # data collator as the wrapped version is only wanted for training.
-        data_collator = getattr(self.eval_data_collator, "__wrapped__", self.data_collator)
+        data_collator = self.eval_data_collator
 
         if isinstance(eval_dataset, torch.utils.data.IterableDataset):
             return poptorch.DataLoader(
@@ -620,9 +620,7 @@ class IPUTrainer:
         if is_datasets_available() and isinstance(test_dataset, datasets.Dataset):
             test_dataset = self._remove_unused_columns(test_dataset, description="test")
 
-        # self.data_collator can be a data collator that was wrapped by pad_on_batch_axis, retrieving the original
-        # data collator as the wrapped version is only wanted for training.
-        data_collator = getattr(self.data_collator, "__wrapped__", self.data_collator)
+        data_collator = self.eval_data_collator
 
         if isinstance(test_dataset, torch.utils.data.IterableDataset):
             return poptorch.DataLoader(
```

jimypbr:
I'm going to merge this and open another PR with my suggested change since Alex doesn't have time to implement this right now.

PR #126, opened 2022-07-19
Add basic documentation

lewtun:
# What does this PR do?

Adds basic documentation that can be rendered on hf.co via https://github.com/huggingface/optimum/pull/311

I've also included a small guide so contributors to this project can understand how the docs should be formatted.

Note: we should remove the bits that checkout the `documentation` branch before merging this PR. That's just used as a test that the approach works end to end (it does :))

Needs https://github.com/huggingface/moon-landing/pull/3521 to be merged before the docs and be viewed from this PR

## Before submitting


.github/workflows/check_doc_build.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: 'huggingface/optimum-graphcore'
+          path: optimum-graphcore
+
+      - name: Make documentation
+        run: |
+          cd optimum-graphcore
+          git fetch origin
```
lewtun, line 26: These 2 lines should be removed before merging

docs/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+# Need node14 to build doc HTML. Taken from https://askubuntu.com/a/1113339
+RUN apt-get update && apt-get install -y \
+    git \
+    curl \
+    gnupg
+RUN curl -sL https://deb.nodesource.com/setup_14.x  | bash -
+RUN apt-get -y install nodejs
+RUN npm install
+
+RUN git clone -b documentation https://github.com/huggingface/optimum-graphcore.git
```
lewtun, line 12: Replace with the following before merging:

```suggestion
RUN git clone https://github.com/huggingface/optimum-graphcore.git
```

.github/workflows/check_doc_build.yml
lewtun, line 26: Remove these 2 lines before merging

docs/Dockerfile
lewtun, line 12: Replace with this before merging:
```suggestion
RUN git clone https://github.com/huggingface/optimum-graphcore.git
```

optimum/graphcore/version.py
```diff
@@ -12,4 +12,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-__version__ = "0.3.1.dev"
+__version__ = "0.3.1.dev0"
```
lewtun, line 15: This is needed to distinguish builds on `main` versus official release branches

tests/test_trainer.py
```diff
@@ -27,15 +27,15 @@
 
 import numpy as np
 
-from huggingface_hub import Repository, delete_repo, login
```
lewtun, line 30: `login` is deprecated, along with using username / password. Refactored to use a dummy token instead

docs/README.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Writing documenation examples
+
+The syntax for Example docstrings can look as follows:
+
+```
+    Example:
+
+    ```python
+    >>> from transformers import Wav2Vec2Processor, Wav2Vec2ForCTC
```
michaelbenayoun, line 270: Maybe put some `optimum-graphcore` code example here?

michaelbenayoun (approved):
Thanks a lot for taking care of this!

docs/README.md
lewtun: Yeah good idea - I'll do that in a separate PR to first get the whole docs built in `optimum` :)

Issue #128, opened 2022-07-25
Test automatic loss scaling and set up option for it


sergiopperez:
Automatic Loss Scaling (ALS) is a feature in the Poplar SDK which brings stability to training large models in half precision, specially when gradient accumulation and reduction across replicas also happen in half precision.

The objective of this issue is to enable the user to turn it on for any model in optimum. Before landing, it's required to test that optimum models converge fine with ALS, in comparison when using a manually tuned static loss scaling.

More details about automatic loss scaling in: https://docs.graphcore.ai/projects/poptorch-user-guide/en/latest/reference.html#poptorch.options._TrainingOptions.setAutomaticLossScaling

PR #133, opened 2022-07-28
Enable --auto-loss-scaling flag in Optimum

sergiopperez:
# What does this PR do?

Enables #128 

## Before submitting

### DIFF
--- a/optimum/graphcore/training_args.py
+++ b/optimum/graphcore/training_args.py
@@ -269,7 +269,9 @@ class IPUTrainingArguments:
     )
     auto_loss_scaling: bool = field(
         default=False,
-        metadata={"help": "Enable automatic loss scaling for half precision training. Note that this is an experimental feature."},
+        metadata={
+            "help": "Enable automatic loss scaling for half precision training. Note that this is an experimental feature."
+        },
     )
     dataloader_mode: str = field(
         default="sync",
--- a/optimum/graphcore/training_args.py
+++ b/optimum/graphcore/training_args.py
@@ -267,6 +267,10 @@ class IPUTrainingArguments:
             "If using automatic loss scaling, this value will be the initial value."
         },
     )
+    auto_loss_scaling: bool = field(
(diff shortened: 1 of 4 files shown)


optimum/graphcore/training_args.py
```diff
@@ -432,6 +436,11 @@ def __post_init__(self):
         else:
             self.gradient_accumulation_steps = 1
 
+        if self.auto_loss_scaling is True:
+            override_str.append(f"auto_loss_scaling={self.auto_loss_scaling}")
+        else:
+            self.auto_loss_scaling = False
```
jimypbr, line 442: this is unnecessary since it's default false

optimum/graphcore/training_args.py
```diff
@@ -432,6 +436,11 @@ def __post_init__(self):
         else:
             self.gradient_accumulation_steps = 1
 
+        if self.auto_loss_scaling is True:
```
jimypbr, line 439: if self.auto_loss_scaling is enough

sergiopperez: Thanks James, so to clarify, are you suggesting to
1) Remove
```
 if self.auto_loss_scaling is True:
      override_str.append(f"auto_loss_scaling={self.auto_loss_scaling}")
else:
      self.auto_loss_scaling = False
```

or 2) Modify it as 
```
if self.auto_loss_scaling:
      override_str.append(f"auto_loss_scaling={self.auto_loss_scaling}")
else:
      self.auto_loss_scaling = False
```

jimypbr: What you did in the commit is right

sergiopperez:
@jimypbr @gejinchen this PR is ready to be reviewed — let me know your thoughts about it. Apart from the commits, I've checked that ALS converges fine for the pretraining of BERT large and GPT2 small.

jimypbr:
Thanks great stuff

PR #139, opened 2022-07-29
Add Groupbert model

ivansche:
# What does this PR do?

This PR includes GroupBert (https://arxiv.org/abs/2106.05822) model, trainable on the IPU. 

Fixes # (issue)

## Before submitting


optimum/graphcore/models/groupbert/modeling_groupbert.py
```diff
@@ -0,0 +1,785 @@
...
+from transformers.modeling_utils import apply_chunking_to_forward
+from transformers.modeling_outputs import BaseModelOutputWithPastAndCrossAttentions
+
+from prettytable import PrettyTable
+
+
+logger = logging.get_logger(__name__)
+
+
+def count_parameters(model):
```
jimypbr, line 59: can we get rid of this? we don't have this for any other model.

jimypbr:
1. I think that you should try to upload pretrained weights to the model hub and get rid of `configs/groupbert-base-uncased.json` and instead use the link to the model hub.
2. Then you should add it the model to `tests/utils.py` so tests are ran on it.

PR #142, opened 2022-08-01
Wav2vec2 Notebooks

thorinf:
# What does this PR do?

Adds 2 Jupyter notebooks; one fine-tuning, and one inference. The inference notebook will use the output from the fine-tuning training.

## Before submitting
- [N] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [N/A] Did you make sure to update the documentation with your changes?
- [N] Did you write any new necessary tests?


notebooks/wav2vec2/requirements.txt
```diff
@@ -0,0 +1,8 @@
+transformers >= 1.18.0
```
jimypbr, line 1: use 4.20.0

jimypbr: also add optimum[graphcore]

jimypbr: I think optimum[graphcore] alone will be sufficent actually. It will get transformers and datasets.

jimypbr:
Really great notebooks 😍

PR #144, opened 2022-08-02
Fix test_pretrained_and_pipelined_models_match for Wav2Vec2

jimypbr:
# What does this PR do?

Fixes tests/test_pipelined_models.py::PipelinedModelsTester for Wav2Vec2 models

## Before submitting
- [N] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [N] Did you make sure to update the documentation with your changes?
- [Y] Did you write any new necessary tests?

### DIFF
--- a/examples/speech-recognition/run_speech_recognition_ctc.py
+++ b/examples/speech-recognition/run_speech_recognition_ctc.py
@@ -629,10 +629,7 @@ def main():
         )
 
         def is_audio_in_length_range(length):
-            try:
-                return length > min_input_length and length < max_input_length
-            except:
-                return False
+            return length > min_input_length and length < max_input_length
 
         # filter data that is shorter than min_input_length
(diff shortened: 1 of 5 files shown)


thorinf (changes requested):
If using the non IPU GS-VQ then the features and the prop-perplexity should the the name of the returned tuple. 

It's a bit confusing because the `tranformers` GS-VQ computes prop-perplexity for training, and code-perplexity for not-training. [But actually both should be computed during both training and validation](https://github.com/facebookresearch/fairseq/blob/4fe8583396191c22011350248119db98ec1b5cb8/fairseq/modules/gumbel_vector_quantizer.py#L161). 

As it stands using the non-IPU quantiser will mean the prop-perplexity is None, which will cause issues when it diversity loss is computed.

PR #154, opened 2022-08-09
Change pad_on_batch_axis collator to repeat examples

jimypbr:
# What does this PR do?

Change `pad_on_batch_axis` to pad by repeating examples from the remainder batch until target size is reached.

When the dataset isn't divisble by the batch size, there is a remainder batch at the end of an epoch. In cases such as QA or small data we need to use all the elements in the dataset, but we need batches to be the same size. We therefore have `pad_on_batch_axis` data collator that pads the remainder batch back up to the right size.

Our previous `pad_on_batch_axis` data collator worked by padding batches to the target size with values that had to be specified by the user. For cross-entropy losses this was nice because we could turn-off the loss on the padded examples by using the special `ignore_index` value as the target label for the padded examples.

This has some drawbacks:
1. Not all losses support a `ignore_index`.
2. It's awkward to use, because the user has to create the data collator and specify the padding values and sizes etc.
3. It's error prone and annoying because you need to know the global batch size and device iterations, however these can be overriden by `ipu_config_overrides` option and the ipu_config isn't updated with those overrides until it is passed into the trainer, after the user has created their padded data collator. So the user will end up putting in the wrong sizes.

This new approach just pads the remainder batch by repeating the examples from the remainder batch. The only small downside is that the last batch in the epoch is bias by seeing the same examples multiple times, but it only happens for a few examples in an epoch so is unlikely to cause problems. The upside is that is makes `pad_on_batch_axis` option much easier to use and task agnostic because the burden is lifted from the user to supply the padding values. It can therefore be entirely moved in the `IPUTrainer` and passed as a option in `IPUTrainingArguments`.

The padding is applied to training, evaluation, and testing in `IPUTrainer`. However for evaluation and testing the outputs of the padded examples are dropped by the trainer, invisibly to the user.

Fixes # (issue)
The issue of having to carefully select the batch size, device iterations for evaluation of QA task so that everything exactly divides. 

## Before submitting


optimum/graphcore/trainer.py
```diff
@@ -204,6 +202,36 @@ def __init__(
...
+        # If no eval_data_collator is specified then use the train data_collator
+        self.eval_data_collator = eval_data_collator if eval_data_collator is not None else self.data_collator
+
+        # If batch axis padding enabled, wrap train/eval data collators with `pad_on_batch_axis` wrapper
+        if self.args.pad_on_batch_axis:
+            if self.args.do_train:
+                logger.info(
+                    "Padding on batch axis enabled, each batch feeded to the compiled model during training will have the proper size"
+                )
+                print("Train bs", self.args.per_device_train_batch_size * self.ipu_config.batch_size_factor())
```
michaelbenayoun, line 216: Is the print intended? If so, we should use `logger.info` instead.

optimum/graphcore/trainer.py
```diff
@@ -204,6 +202,36 @@ def __init__(
...
+                data_collator_wrapper = pad_on_batch_axis(
+                    self.args.per_device_train_batch_size * self.ipu_config.batch_size_factor()
+                )
+                self.data_collator = data_collator_wrapper(data_collator)
+
+            if self.args.do_eval:
+                logger.info(
+                    "Padding on batch axis enabled, each batch feeded to the compiled model during training will have the proper size"
+                )
+                print(
```
michaelbenayoun, line 226: Same comment.

michaelbenayoun (approved):
LGTM, that's really awesome James!
Do you think we should enable `pad_on_batch_axis` by default?

optimum/graphcore/trainer.py
jimypbr: Nope. Debugging stuff left in.

optimum/graphcore/trainer.py
jimypbr: Nope. Debugging stuff left in.

jimypbr:
> LGTM, that's really awesome James! Do you think we should enable `pad_on_batch_axis` by default?

:) Yes probably. Let's leave it off and merge and see how my PR #130 gets on with it on first.

PR #162, opened 2022-08-10
Doc build requires Python 3.7 or higher

regisss:
# What does this PR do?

The new setup constraint `transformers==4.20.0` makes `python>=3.7.0` mandatory. However, the Docker image used for building the documentation is based on the image `graphcore/pytorch:latest`, which may be based on Ubuntu 18.04 that has Python 3.6 as the default Python 3 version.

~This PR adds a step in the Dockerfile to install Python 3.8 in order to solve this.~ This does not work because PopART will complain that it was built against Python 3.6.

This PR replaces the tag `latest` of the base Docker image used in `docs/Dockerfile` with the tag `2.6.0-ubuntu-20.04`. That means the tag will have to be updated for each new version of the SDK.

## Before submitting


michaelbenayoun:
@jimypbr Is any docker image based on Python 3.7+ planned?

jimypbr:
`docker run graphcore/pytorch:2.6.0-ubuntu-20.04 bash -c "pip install numpy; python -c 'import popart'"` does this run with error for you?

regisss:
> The ubuntu 20.04 docker image is Python 3.8 so this should work. `docker run graphcore/pytorch:2.6.0-ubuntu-20.04 bash -c "pip install numpy; python -c 'import popart'"` does this run with error for you?

Yes it does work. The issue is with the current image `graphcore/pytorch:latest` which currently points to `graphcore/pytorch:2.6.0-ubuntu-18.04`. This one does not support Python 3.7+.

jimypbr:
Yes agreed! Go with `pytorch:2.6.0-ubuntu-20.04`

regisss:
Cool, that's the current change of this PR! So I think we can merge it as it is @michaelbenayoun ?

jimypbr:
failing test case fixed on `main` branch FYI

michaelbenayoun:
Yes!

PR #164, opened 2022-08-12
Added multiple choice notebook

hja03:


jimypbr:
call `trainer.evaluate()`

PR #167, opened 2022-08-17
Generate PR doc from forks

regisss:
# What does this PR do?

The PR doc generation would fail if the PR comes from a fork. This PR fixes that by extracting the clone URL in the GH workflow and providing it to the Dockerfile.

It also adds a default clone URL to the Makefile so that the main documentation of Optimum (the parent package) can be generated without any optimum-graphcore-PR clone URL.

## Before submitting


lewtun (approved):
Thanks for the fix @regisss - LGTM 🔥 !

PR #173, opened 2022-08-31
Run each test-example completely with dataset in a tempdir 

### LINKED ISSUE #172: Fix test-examples CI
- test-examples is failing on the last test `convnext`.
- I suspect that it's because there it is running out of space on the CI machine and so the last test fails due to this.
- If we isolate all the tests so the datasets and running is done inside of a temporary directory then there should be no side-effects and this hopefully fixes the issue.

jimypbr:
# What does this PR do?

* Run each example in `test_examples.py` in a temp directory in the examples directory.
* Also set the `HF_HOME` to that examples directory. This way the datasets/caches are created and destroyed with every test case, leaving nothing behind that could cause side effects.

This should fix #172

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -39,9 +39,6 @@ from transformers.testing_utils import slow
 from .utils import MODELS_TO_TEST_MAPPING
 
 
-_ALLOWED_REPLICATION_FACTOR = 2
-
-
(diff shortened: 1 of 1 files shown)


michaelbenayoun (approved):
Great!
Did you run it once?

PR #177, opened 2022-09-06
Compute max_num_masked_tokens dynamically

gejinchen:
## Before submitting

### DIFF
--- a/examples/language-modeling/run_pretraining.py
+++ b/examples/language-modeling/run_pretraining.py
@@ -163,13 +163,6 @@ class DataTrainingArguments:
             "skipped."
         },
     )
-    is_already_preprocessed: Optional[bool] = field(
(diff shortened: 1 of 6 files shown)


examples/language-modeling/run_pretraining.py
```diff
@@ -163,13 +163,6 @@ class DataTrainingArguments:
             "skipped."
         },
     )
-    is_already_preprocessed: Optional[bool] = field(
```
jimypbr, line 166: why are you deleting this? this is still needed, I think

gejinchen: It's duplicated. There is a same one right above it.

jimypbr: Fair enough!

PR #181, opened 2022-09-20
Add DistilBERT

katalinic-gc:
# What does this PR do?

Add DistilBERT for upstream tasks of Masked LM, Sequence Classification, Token Classification, Question Answering and Multiple Choice. A corresponding model card with the default IPU config for tests has been created separately. The examples test for Multiple Choice yields a eval score below the threshold of 0.75, so a per-model override is added to allow the test to pass. 

## Before submitting


optimum/graphcore/models/distilbert/modeling_distilbert.py
```diff
@@ -0,0 +1,311 @@
...
+    register,
+)
+
+
+logger = logging.get_logger(__name__)
+
+FLOAT16_LIMIT = 1e4
+
+
+class IPUMultiHeadSelfAttention(MultiHeadSelfAttention):
```
michaelbenayoun, line 49: This is needed only for the float 16 mask value? Just asking to know as it soon won't be needed anymore.

tests/test_examples.py
```diff
@@ -383,6 +389,7 @@ class MultipleChoiceExampleTester(ExampleTesterBase, metaclass=ExampleTestMeta,
     # Using a small gradient accumulation steps value because input data is repated for the multiple choice task.
     TRAIN_BATCH_SIZE = 1
     EVAL_BATCH_SIZE = 1
+    EVAL_SCORE_THRESHOLD_OVERRIDES = {"distilbert-base-uncased": 0.65}
```
michaelbenayoun, line 392: That's great but I would use the model type (e.g. `distilbert`) instead of the model name here, this makes it less specific since model names are specified elsewhere.

michaelbenayoun (approved):
Awesome addition, thanks!

tests/test_examples.py
katalinic-gc: I am under the impression there's not a trivial way to go from model name to model type, so some possibilities I see are

- pass `model_type` to `_create_test` in addition to `model_name` and `ipu_config_name`
- within `_create_test` infer `model_type` by looping through `MODELS_TO_TEST_MAPPING` to find the relevant key
- within `_create_test` get the auto config from the model name, and retrieve model type from that

Could you please advise on what you think would be most appropriate?

michaelbenayoun: You're right, let's keep it like that for now!

katalinic-gc: Thanks for confirming

gejinchen (approved):

PR #190, opened 2022-10-10
[WIP] Pipelines

gejinchen:
1. Support inference pipelines.
2. Currently, the supported tasks are `audio-classification`, `automatic-speech-recognition`, `fill-mask`, `image-classification`, `question-answering`, `text-classification`, `token-classification` and `zero-shot-classification`.

## Before submitting


.github/workflows/test-general.yml
```diff
@@ -38,4 +38,4 @@ jobs:
         export SDK_PATH=/opt/gc/poplar_sdk-ubuntu_20_04-3.0*
         . ${SDK_PATH}/poplar-ubuntu_20_04*/enable.sh
         . ${SDK_PATH}/popart-ubuntu_20_04*/enable.sh
-        pytest tests/
+        pytest tests/ --ignore=tests/pipelines/
```
michaelbenayoun, line 41: How long until we cover those?

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+from .zero_shot_classification import IPUZeroShotClassificationPipeline
+
+
+logger = logging.get_logger(__name__)
+
+TASK_ALIASES = {
+    "sentiment-analysis": "text-classification",
+    "ner": "token-classification",
+}
+SUPPORTED_TASKS = {
```
michaelbenayoun, line 43: Is all this copy / paste needed to provide the default ipu config?
I mean for those not changing the `impl` class.

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+        "class": (AutoModelForAudioClassification,),
+        "default": {
+            "model": ("superb/hubert-base-superb-ks", "d7e0efe"),
+            "ipu_config": "Graphcore/hubert-base-ipu",
+        },
+        "type": "audio",
+    },
+    "automatic-speech-recognition": {
+        "impl": AutomaticSpeechRecognitionPipeline,
+        # "class": (AutoModelForCTC, AutoModelForSpeechSeq2Seq),
```
michaelbenayoun, line 55: Remove?

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+    opts.setExecutionStrategy(poptorch.ShardedExecution())
+    model = poptorch.inferenceModel(model.eval(), opts)
+    return model
+
+
+def get_inference_context(self):
+    return torch.no_grad
+
+
+def check_model_type(self, supported_models: Union[List[str], dict]):
```
michaelbenayoun, line 158: Consistent type annotation, I guess to match the rest of the library:
```suggestion
def check_model_type(self, supported_models: Union[List[str], Dict[Type, Type]]):
```
(just guessing here for the "`Type`" part.

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+        supported_models_names = []
+        for config, model in supported_models.items():
+            # Mapping can now contain tuples of models for the same configuration.
+            if isinstance(model, tuple):
+                supported_models_names.extend([_model.__name__ for _model in model])
+            else:
+                supported_models_names.append(model.__name__)
+        supported_models = supported_models_names
+    if self.model._user_model.__class__.__bases__[0].__name__ not in supported_models:
+        logger.error(
```
michaelbenayoun, line 176: Should we raise something here?

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+                f"{model} is not a valid model_id."
+            )
+        task = get_task(model, use_auth_token)
+
+    if task in TASK_ALIASES:
+        task = TASK_ALIASES[task]
+
+    targeted_task = "translation" if task.startswith("translation") else task
+
+    if targeted_task not in list(SUPPORTED_TASKS.keys()):
```
michaelbenayoun, line 215: nit: no need to cast to list, and to get the keys either.

optimum/graphcore/pipelines/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+        # the model on the other hand might have a tokenizer, but
+        # the files could be missing from the hub, instead of failing
+        # on such repos, we just force to not load it.
+        load_tokenizer = False
+    else:
+        load_tokenizer = True
+    if targeted_task in NO_FEATURE_EXTRACTOR_TASKS:
+        load_feature_extractor = False
+    else:
+        load_feature_extractor = True
```
michaelbenayoun, line 230:
```suggestion
    # These will never require a tokenizer.
    # the model on the other hand might have a tokenizer, but
    # the files could be missing from the hub, instead of failing
    # on such repos, we just force to not load it.
    load_tokenizer = targeted_task not in NO_TOKENIZER_TASKS
    load_feature_extractor = targeted_task not in NO_FEATURE_EXTRACTOR_TASKS
```

tests/pipelines/test_pipelines_audio_classification.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2021 The HuggingFace Team. All rights reserved.
```
michaelbenayoun, line 1: Maybe add this copyright thing in `tests/pipelines/__init__.py`

tests/pipelines/test_pipelines_audio_classification.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertEqual(
+            output,
+            [
+                {"score": ANY(float), "label": ANY(str)},
+                {"score": ANY(float), "label": ANY(str)},
+            ],
+        )
+
+    @require_torch
+    def test_small_model_pt(self):
```
michaelbenayoun, line 82: Why?

michaelbenayoun (approved):
Huge work, thanks @gejinchen !!
I guess it is okay to merge with all those commented tests, but we should make sure to handle them soon.

.github/workflows/test-general.yml
gejinchen: 1. Most of the tasks' tests could work, except the commented ones, which could take a while to uncomment since we simply don't support some of those models.
2. `automatic-speech-recognition` is a bit special so may need some refactoring to support it.
3. The other concern is that these tests are time-consuming

optimum/graphcore/pipelines/__init__.py
gejinchen: It's not just about providing the default ipu config. I also removed all the redundant TensorFlow stuff from the old `SUPPORTED_TASKS `. Also, I think this is a good place to see exactly which tasks are supported.

jimypbr: There's also default arguments in here like `padding_length`.

tests/pipelines/test_pipelines_audio_classification.py
gejinchen: If you mean why comment this out, it's because `Wav2Vec2ForSequenceClassification` is not supported yet.

optimum/graphcore/pipelines/__init__.py
gejinchen: I think this is a reasonable change. However, my `check_model_type` is just a function override, so I think it is wise to keep the change minimal and leave the other changes to the `transformers` library.

PR #197, opened 2022-11-09
Add a notebook for the stable diffusion pipeline

katalinic-gc:
# What does this PR do?

Adds an example notebook running the text-to-image Stable Diffusion pipeline on the IPU.

## Before submitting


notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+
+from diffusers import StableDiffusionPipeline, UNet2DConditionModel
+from optimum.graphcore import IPUConfig
+from optimum.graphcore.modeling_utils import PipelineMixin
+import torch
+
+import poptorch
+
+
+def _sliced_attention(self, query, key, value, sequence_length, dim):
```
payoto, line 13: Can you explain why this is needed (in the docstring), are those memory optimisation to fit on 4 IPUs?

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+            if module is not None:
+                module = module.float()
+                kwargs[module_name] = module
+
+        unet = kwargs.get("unet", None)
+        if unet is not None:
+            IPU_CONFIG_DICT = {
+                "inference_device_iterations": 1,
+                "inference_replication_factor": {"default": 1},
+                "executable_cache_dir": "./exe_cache",
```
payoto, line 116: Can this be customised when the model is loaded or is the fact that it is hard coded here make that impossible? We're going to need the ability to customise it.

notebooks/stable_diffusion/ipu_models.py
katalinic-gc: The implementation of the function itself is being over-riden because we currently don't support slice assignment, hence we are resorting to concatenation. Until it's implemented, can't say how the memory profile would compare otherwise.

notebooks/stable_diffusion/ipu_models.py
katalinic-gc: I'll look into what can be done here.

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+    # reshape hidden_states
+    hidden_states = self.reshape_batch_dim_to_heads(hidden_states)
+    return hidden_states
+
+
+def override_sliced_attention(unet):
+    for module in unet.modules():
+        if module.__class__.__name__ == "CrossAttention":
+            assert hasattr(module, "_sliced_attention")
+            setattr(module, "_sliced_attention", partial(_sliced_attention, module))
```
michaelbenayoun, line 48:
```suggestion
            module._sliced_attention = _sliced_attention.__get__(module, CrossAttention)
```

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+def copy_signature(base):
+    def decorator(func):
+        func.__signature__ = inspect.signature(base)
+        return func
+
+    return decorator
+
+
+class IPUStableDiffusionPipeline(StableDiffusionPipeline):
+    @copy_signature(StableDiffusionPipeline.__init__)
```
michaelbenayoun, line 103: `functools.wraps` might be doing what you're trying to do.

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+                "enable_half_partials": True,
+            }
+            ipu_config = IPUConfig.from_dict(IPU_CONFIG_DICT)
+            unet_ipu = copy.deepcopy(unet)
+            unet_ipu.__class__ = IPUUNet2DConditionModel
+            override_sliced_attention(unet_ipu)
+            unet_ipu.ipu_config = ipu_config
+            unet_ipu.parallelize()
+            opts = ipu_config.to_options(for_inference=True)
+            opts.setExecutionStrategy(poptorch.ShardedExecution())
```
michaelbenayoun, line 128: I think that's already the case

jimypbr: yes this is already the default for inference

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+    hidden_states = torch.cat(hidden_states)
+
+    # reshape hidden_states
+    hidden_states = self.reshape_batch_dim_to_heads(hidden_states)
+    return hidden_states
+
+
+def override_sliced_attention(unet):
+    for module in unet.modules():
+        if module.__class__.__name__ == "CrossAttention":
```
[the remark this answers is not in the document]
katalinic-gc: Does this and comment below not require the importing of CrossAttention? (which I believe diffusers does not expose)

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,137 @@
...
+            assert hasattr(module, "_sliced_attention")
```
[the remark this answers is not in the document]
katalinic-gc: Removed

notebooks/stable_diffusion/ipu_models.py
katalinic-gc: Made the ipu_config passable as an arg

notebooks/stable_diffusion/ipu_models.py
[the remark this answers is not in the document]
michaelbenayoun: You can still import it using the full path.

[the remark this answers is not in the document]
katalinic-gc: Indeed, fixed :)

notebooks/stable_diffusion/ipu_models.py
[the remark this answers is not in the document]
katalinic-gc: I was mistaken, updated per Michael's suggestions

notebooks/stable_diffusion/ipu_models.py
payoto: I don't see the code path that leads to it? If it's the case could you make the argument explicit?

For the work we have planned I only really need to be able to set the path of the executable cache directory, but Ideally we might want to be able to override any of the defaults without touching the others. Something along the lines of:

```python
def __init__(ipu_config: Optional[Dict]=None, **kwargs):
    ...
    IPU_CONFIG_DICT = {...} # the one you have set
    if ipu_config:
        IPU_CONFIG_DICT.update(ipu_config)
```

What do you think of that?

payoto (commented):
Just looking for the clarification on how to set the config options (and maybe making it easier)

notebooks/stable_diffusion/ipu_models.py
katalinic-gc: > I don't see the code path that leads to it? If it's the case could you make the argument explicit?

The initialisation is handled by `from_pretrained`, which will go through args and kwargs of `__init__`, load whatever it recognises, and pass them to `__init__`.

> What do you think of that?

Agreed, will change it to update an existing default.

katalinic-gc: Updated to basically follow your suggestion

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,149 @@
...
+
+def maybe_cast_module_to_float(module):
+    if module is not None:
+        module = module.float()
+    return module
+
+
+class IPUStableDiffusionPipeline(StableDiffusionPipeline):
+    def __init__(
+        self, vae, text_encoder, tokenizer, unet, scheduler, safety_checker, feature_extractor, ipu_config=None
```
payoto, line 104:
```suggestion
        self, vae, text_encoder, tokenizer, unet, scheduler, safety_checker, feature_extractor, ipu_config=None, **kwargs
```

notebooks/stable_diffusion/ipu_models.py
```diff
@@ -0,0 +1,149 @@
...
+            unet = UNetCastingWrapper(unet_ipu)
+
+        super().__init__(
+            vae=vae,
+            text_encoder=text_encoder,
+            tokenizer=tokenizer,
+            unet=unet,
+            scheduler=scheduler,
+            safety_checker=safety_checker,
+            feature_extractor=feature_extractor,
```
payoto, line 144:
```suggestion
            feature_extractor=feature_extractor,
            **kwargs,
```

payoto (approved):
Can you just keep the `**kwargs`? That's to make sure that if the interface of the model we inherit from changes we don't need to edit our implementation every time.

Otherwise LGTM

katalinic-gc:
> Can you just keep the `**kwargs`? That's to make sure that if the interface of the model we inherit from changes we don't need to edit our implementation every time.
> 
> Otherwise LGTM

Hmm adding `**kwargs` to `__init__` doesn't seem that straightforward. It seems the implementation of `from_pretrained` from https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipeline_utils.py#L513 onwards throws an error complaining about missing modules as it interprets kwargs as an additional entry from the signature.

notebooks/stable_diffusion/ipu_models.py
payoto: Not applicable see later comment

notebooks/stable_diffusion/ipu_models.py
payoto: Not applicable see later comment

payoto:
Ok, in that case it seems like you've got the best solution available. 
LGTM

PR #205, opened 2022-11-22
Use correct package name for scikit-learn in requirements

lesteve:
# What does this PR do?

It uses `scikit-learn` rather than `sklearn` in a `requirements.txt` file.

scikit-learn is the actual package name, using `sklearn` in requirement files will stop working December 1st 2022. For more details, see https://github.com/scikit-learn/sklearn-pypi-package.

## Before submitting

I checked the small issue box since it seems to apply best.


jimypbr (approved):
LGTM thanks for fixing!

PR #214, opened 2022-11-29
Notebook-specific cache dirs

rahult-graphcore:
Change 

- Each notebook to have its own poplar executable cache directory. This is needed as a workaround to prevent long .popef file loading times over cloud networks, to ensure fast compilation times.


payoto (approved):
LGTM, for a bit of context to other reviewers, we need this to work around slow load times when you have folders with a lot of Poplar executables.

jimypbr (approved):
LGTM

PR #218, opened 2022-12-05
Update wav2vec2 reqs

rahult-graphcore:
Changes
- Install `optimum-graphcore` instead of `optimum[graphcore] as we've seen cases where the latter installs an older version
- Remove installing an older transformers version


payoto (approved):
LGTM thanks @rahult-graphcore

jimypbr (approved):

PR #219, opened 2022-12-06
Sentiment analysis using BERT class models

payoto:
This PR adds:

- A notebook showing the use of pipelines for sentiment analysis
- A notebook showing how to manage IPUs from notebooks
- ~Some small tweaks to the pipeline interface to make it easier to manage IPU configs.~
- Addition of a `list_tasks` method to `pipelines`


optimum/graphcore/pipelines/__init__.py
```diff
@@ -217,12 +227,32 @@ def pipeline(
     tokenizer: Optional[Union[str, PreTrainedTokenizer]] = None,
     feature_extractor: Optional[Union[str, PreTrainedFeatureExtractor]] = None,
     revision: Optional[str] = None,
-    use_fast: bool = True,
     use_auth_token: Optional[Union[str, bool]] = None,
     pipeline_class: Optional[Any] = None,
     fp16: bool = True,
+    ipu_config_kwargs: Optional[Dict[str, Any]] = None,
```
jimypbr, line 233: why do we need ipu config kwargs? You can just just modify the `ipu_config`.

payoto: I have to create a whole new config, not modify an existing one. I guess I don't have a good feel for how reusable/throw-away these configs should be?

payoto: Ok removed and used the ipu_config argument

jimypbr (approved):

PR #227, opened 2022-12-09
wav2vec2 inference to use cachedir

rahult-graphcore:
Update the Wav2vec2 - inference notebook to use the poplar exe cache directory


payoto (commented):
LGTM

PR #238, opened 2022-12-16
detach from IPUs to free resources

rahult-graphcore:
We add more explicit "detachFromDevice()" calls to free up IPU resources so models in upcoming cells have IPUs to attach to.


payoto (approved):
LGTM

Issue #198, opened 2022-11-11
Run on paperspace buttons still point to 2.6 SDK


jimypbr:
- Need updating to SDK 3.0

PR #239, opened 2022-12-20
Make compatible with SDK3.1

payoto:
Updates requirements to make Optimum Graphcore compatible with PopTorch in SDK 3.1 and torch 1.13


jimypbr:
This also fixes #198

jimypbr (approved):

PR #241, opened 2023-01-06
PackedBERT SQuAD fine-tuning notebook

arsalanu:
# What does this PR do?

Adds a variant of the notebook question_answering with a step-by-step implementation of packed-BERT for fine-tuning. There are modifications to preprocessing, dataset creation, the model class and postprocessing in comparison to unpacked question-answering, all highlighted in this notebook. I've read through and its ready for review, there could be stuff to add or remove. 

## Notes on the notebook

Question: We would ideally like to have interleaved training-validation loops, but SQuAD validation accuracy metrics require custom processing for logits with multiple arguments to the `postprocess_qa_predictions` function. `IPUTrainingArguments` lets you add a `preprocess_logits_for_metrics` function but this only takes `logits` as an argument, with no discernible way to add multiple arguments or even kwargs (let me know if this is possible). For this reason I've done the validation process separately (after training) using `.predict()`. 

I've also added an example showing inference using HF pipelines with a single text input, because of the preset task-specific pipelines I've used an unpacked input with the default 'question answering' pipeline to prove model functionality. We can't yet do packed inference without incorporating preprocessing into the inference pipeline. I will remove this if it's not needed.


payoto:
Can you rebase on main please?

PR #243, opened 2023-01-10
Remove colab links

rahult-graphcore:
This commit removes links to colab notebooks (inherited from upstream notebooks) from optimum notebooks as we received feedback from paperspace that it would be confusing to the user.


payoto (approved):
LGTM

PR #246, opened 2023-01-12
Update links to gradient for SDK 3.1

payoto:
We have released a new version of the SDK and need to update those links as the old ones point to a version which will not work.


jimypbr (approved):
Looks good. Tested all the links and they work.

PR #250, opened 2023-01-23
Remove deprecated IPU config arguments

katalinic-gc:
# What does this PR do?

Fixes # (issue)

## Before submitting


jimypbr (commented):
Actually just delete `examples/language-modeling/ipu_config.json` entirely. It's some legacy thing that's not used anymore.

jimypbr:
I think we'll need to go on HF hub and delete these options from all the IPU configs when we land this

PR #262, opened 2023-02-09
Add Whisper model for inference

paolot-gc:
# What does this PR do?

This adds Whisper inference support to optimum-graphcore.
At this stage, support is provided as proof-of-concept only. 
A Jupyter Notebook demonstrates how to use Whisper on some audio data, based on the example on the original HF Whisper README page.
In principle, larger versions of model can be accommodated (see the `model_size` variable in the notebook).

## Before submitting


optimum/graphcore/generation_utils.py
```diff
@@ -63,31 +86,22 @@ def _prepare_encoder_decoder_kwargs_for_generation(
...
         encoder_kwargs[model_input_name] = inputs_tensor
-        model_kwargs["encoder_outputs"]: ModelOutput = encoder(**encoder_kwargs)
 
-        # 5. undo the casting for cpu
-        if dtype is torch.float16:
-            encoder.to(torch.float16)
-            for key, output in model_kwargs["encoder_outputs"].items():
-                if isinstance(output, torch.Tensor) and output.dtype is torch.float32:
-                    model_kwargs["encoder_outputs"][key] = output.to(torch.float16)
+        if not hasattr(self, "poptorch_encoder"):
```
jimypbr, line 102: remove this for now. I want to fully support this feature for all models.

optimum/graphcore/generation_utils.py
```diff
@@ -223,6 +237,7 @@ def greedy_search(
 
         # Change: disable use_cache because it can't be statically compiled
         if "use_cache" in model_kwargs:
+            # print("Overriding use_cache setting... - use_cache=False")          # TODO Find an obtrusive way to let the user know about the override
```
jimypbr, line 240: use `logger.warn` for this message

optimum/graphcore/generation_utils.py
```diff
@@ -496,6 +506,7 @@ def beam_search(
 
         # Change: disable use_cache because it can't be statically compiled
         if "use_cache" in model_kwargs:
+            # print("Overriding use_cache setting... - use_cache=False")    # TODO Find a way to let the user know
```
jimypbr, line 509: again use `logger.warn`

optimum/graphcore/generation_utils.py
```diff
@@ -797,6 +810,7 @@ def sample(
 
         # Change: disable use_cache because it can't be statically compiled
         if "use_cache" in model_kwargs:
+            # print("Overriding use_cache setting... - use_cache=False")   # TODO Find a way to let the user know
```
jimypbr, line 813: `logger.warn`

optimum/graphcore/generation_utils.py
```diff
@@ -515,6 +526,8 @@ def beam_search(
 
             model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
 
+            model_inputs["decoder_input_ids"] = model_inputs["decoder_input_ids"][:,-1:]     # PT
```
jimypbr, line 529: Is this whisper specific? will it break other models?

optimum/graphcore/models/whisper/modeling_whisper.py
```diff
@@ -0,0 +1,608 @@
...
+        hidden_states = self.activation_fn(self.fc1(hidden_states))
+        hidden_states = nn.functional.dropout(hidden_states, p=self.activation_dropout, training=self.training)
+        hidden_states = self.fc2(hidden_states)
+        hidden_states = nn.functional.dropout(hidden_states, p=self.dropout, training=self.training)
+        hidden_states = residual + hidden_states
+
+        # NOTE: We never clamp. This differs from the original implementation
+        # With the code below, compilation fails with an error
+        # 023-02-08T10:24:33.204593Z popart:popart 3749453.3749453 E: np broadcasting failed on Op 197 (ai.onnx.Add:7), incompatible types FLOAT and FLOAT16 (shapes [1500 384] and [384]) 
+
```
jimypbr, line 129: Can you add remove the error message and just say:
```
# NOTE: This differs from the original implementation
# There is a type mismatch bug with this call to clamp so we remove it here. It is anyway not need on IPU because FP16 values are clamped to maximum value by default.
# TODO: when bug is fixed in future SDK remove this entire class;
```

optimum/graphcore/models/whisper/modeling_whisper.py
```diff
@@ -0,0 +1,608 @@
...
+            input_shape = input_ids.size()
+            input_ids = input_ids.view(-1, input_shape[-1])
+        elif inputs_embeds is not None:
+            input_shape = inputs_embeds.size()[:-1]
+        else:
+            raise ValueError("You have to specify either decoder_input_ids or decoder_inputs_embeds")
+
+        # past_key_values_length
+        past_key_values_length = past_key_values[0][0].shape[2] if past_key_values is not None else 0     # PT <<<<<
+        # past_key_values_length = past_key_values[0][2] if past_key_values is not None else 0     # PT Change
```
jimypbr, line 374: These look like works in progress changes. Can you revert them back to the original, please?

optimum/graphcore/generation_utils.py
paolot-gc: Removed and it still works. Not sure why I added it.

optimum/graphcore/models/whisper/modeling_whisper.py
paolot-gc: Done

optimum/graphcore/models/whisper/modeling_whisper.py
paolot-gc: Done

optimum/graphcore/generation_utils.py
paolot-gc: Please clarify; remove exactly what?

optimum/graphcore/generation_utils.py
jimypbr: This is warning if `use_cache` is just in the model_kwargs which is always is. It should only warn if the user has passed `use_cache=True`

optimum/graphcore/models/whisper/modeling_whisper.py
```diff
@@ -0,0 +1,617 @@
...
+        encoder_outputs: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
+        past_key_values: Optional[Tuple[Tuple[torch.FloatTensor]]] = None,
+        decoder_inputs_embeds: Optional[Tuple[torch.FloatTensor]] = None,
+        labels: Optional[torch.LongTensor] = None,
+        use_cache: Optional[bool] = None,
+        output_attentions: Optional[bool] = None,
+        output_hidden_states: Optional[bool] = None,
+        return_dict: Optional[bool] = None,
+    ) -> Union[Tuple[torch.Tensor], Seq2SeqModelOutput]:
+        # when encoder_outputs is passed as BaseModelOutput, it comes out as a dict...
```
jimypbr, line 594: Can you remove this and try with SDK3.2?

PR #266, opened 2023-02-17
Packed BERT notebooks and separate models/utils updated to latest

arsalanu:
# What does this PR do?
* Simplified notebooks for all three supported Packed BERT tasks for easy implementation
* Adds all of the necessary utils/model heads imported into notebooks - preprocessing, postprocessing, model changes

**Notes:** For the time being, the  `models/` and `utils/` are in this folder that goes into `notebooks/` but ideally it would be nice to have the utils put into `optimum/graphcore/` so they could be easily importable with the package - and the `models/modeling_bert_packed.py` could just be options within the default `modeling_bert`, and packing could be enabled through the AutoConfig (some tweaks would be needed for that, but nothing extensive) This also gives us a structure to add future packing tasks/notebooks

## Fixes
* I've removed the model classes and packing algorithm/dataset creation utils from the notebooks, noted that they were too complex and large as notebook code blocks requiring too much explanation and would be hard to maintain here. The intention of these notebooks is to give brief explanations of the differences between unpacked and packed at each stage and allows users to easily implement it using the importable methods.
* A more in depth explanation of the packing/preproc/postproc/model change process will get its own notebooks/blog in future so we don't need to cover it for this notebook
* I've used the env variables for pod type and executable dir
* Rewritten most of these notebooks to not be as detailed/complex and use more active language - some of it is copied from existing notebooks for the same (unpacked) tasks - happy to change stuff


notebooks/packed_bert/models/modeling_bert_packed.py
```diff
@@ -0,0 +1,282 @@
...
+        return self.output(
+            loss= total_loss,
+            start_logits= unpacked_start_logits, 
+            end_logits=unpacked_end_logits, 
+            hidden_states=final_layer_output.hidden_states, 
+            attentions=final_layer_output.attentions
+        )
+
+
+class PackedPipelinedBertForQuestionAnswering(BertForQuestionAnswering, BertPipelineMixin):
```
jimypbr, line 229: The naming convention we have is `PipelinedPackedBert...`

notebooks/packed_bert/models/modeling_bert_packed.py
```diff
@@ -0,0 +1,282 @@
...
+
+            logits = logits.reshape([batch_dim, self.max_seq_per_pack, logits.shape[-1]])
+
+            return (loss, logits)
+        else:
+            return logits
+
+
+
+class PackedPipelinedBertForSequenceClassification(BertForSequenceClassification, BertPipelineMixin):
```
jimypbr, line 103: `PipelinedPacked...`

jimypbr (approved):

PR #270, opened 2023-02-23
set pod_type in tests rather than replication factor

rahult-graphcore:
Sets the pod_type explicitly rather than the replication factor. Prefer to have the replication factor implied, because not all models have the same `ipus_per_replica`.


jimypbr (approved):
LGTM

PR #271, opened 2023-02-27
Pod type sets replication factor

rahult-graphcore:
- `replication_factor` and `inference_replication_factor` will be set by `pod_type`, if provided. Can be overruled by `ipu_config_overrides`. 
- Also includes a warning that dict values in the ipu_config.json will be deprecated in the next release.


jimypbr (changes requested):
There are several ipu_configuration tests related to `pod_type` that will need to be updated or removed by this change.

jimypbr:
smallest pod type is pod4 but we now have things that use less than 4 IPUs. We should add pod1 and pod2 to the allowed values, or?

jimypbr:
Yeah that's the simplest. `num_ipus` or something. Deprecate `pod_type` and recommend `num_ipus`.

jimypbr:
Can you keep pod_type argument but add a deprecation warning telling people to use n_ipu instead please? Need to maintain backwards compatibility for a version.

jimypbr:
Can you update the `test_examples.py` to use n_ipu arg?

optimum/graphcore/training_args.py
```diff
@@ -736,6 +740,14 @@ def __post_init__(self):
                 FutureWarning,
             )
 
+        if self.pod_type is not None:
+            warnings.warn(
+                "`pod_type` is deprecated and will be removed in the next release of Optimum Graphcore. Use `n_ipu` "
+                "instead tp specify how many IPUs you would like the Trainer to use.",
```
jimypbr, line 746: typo: tp->to

jimypbr (approved):

PR #272, opened 2023-02-27
Fix most pipeline tests

katalinic-gc:
# What does this PR do?

Fixes up most of the pipeline tests we would reasonably expect to pass.

## Before submitting


optimum/graphcore/pipelines/__init__.py
```diff
@@ -224,6 +227,8 @@ def get_poplar_executor(
     ipu_config.inference_device_iterations = 1
     # TODO: inference_replication_factor should be adaptive, especially for batching.
     ipu_config.inference_replication_factor = 1
+    if not fp16:
+        ipu_config.enable_half_partials = False
```
jimypbr, line 231: good catch

optimum/graphcore/pipelines/__init__.py
```diff
@@ -398,6 +398,21 @@ def pipeline(
     old_forward = pipeline_class._forward
 
     def new_forward(self, model_inputs, *args, **kwargs):
+        if isinstance(self.model, poptorch.PoplarExecutor) and targeted_task not in SUPPORTED_GENERATION_TASKS:
```
jimypbr, line 401: I thought this would support text-gen + beam search?

jimypbr (commented):
How long do the tests take? Does this PR enable them for pull request CI?

katalinic-gc:
> How long do the tests take? Does this PR enable them for pull request CI? It's probably too long for that, right?

4 hrs with slow tests being skipped.

optimum/graphcore/pipelines/__init__.py
katalinic-gc: I was intending to only support it for non text-gen pipelines as we did before text gen landed. If we want to support it for text gen as well, I can see if it works given you found it kept recompiling.

jimypbr:
I think you need to rebase to main

.github/workflows/test-pipelines.yml
```diff
@@ -0,0 +1,43 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Create and start a virtual environment
+      run: |
+        python -m venv base
+    - name: Install dependencies
+      run: |
+        source base/bin/activate
+        pip install --upgrade pip
+        export SDK_PATH=/opt/gc/poplar_sdk-ubuntu_20_04-3.1*
```
jimypbr, line 31: change to 3.2

PR #273, opened 2023-02-27
Efficient decoder text generation wrapper

jimypbr:
# What does this PR do?

Current text generation returns all the logits back from IPU to host. However we only need the logits from the token to be generated. This wraps the model to return only what is necessary to improve the IO performance of the text generation.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/pipelines/__init__.py
+++ b/optimum/graphcore/pipelines/__init__.py
@@ -413,8 +413,6 @@ def pipeline(
         # For text generation models, deallocate the internal poplar executor
         if hasattr(self.model, "poptorch_decoder"):
             self.model.poptorch_decoder.destroy()
-        if hasattr(self.model, "poptorch_encoder"):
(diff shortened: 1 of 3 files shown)


optimum/graphcore/pipelines/__init__.py
```diff
@@ -411,8 +411,10 @@ def new_forward(self, model_inputs, *args, **kwargs):
     # Implement pipelines __del__ to clean up poplar exector
     def _del(self):
         # For text generation models, deallocate the internal poplar executor
-        if hasattr(self.model, "poptorch_model"):
-            self.model.poptorch_model.destroy()
+        if hasattr(self.model, "poptorch_decoder"):
+            self.model.poptorch_decoder.destroy()
+        if hasattr(self.model, "poptorch_encoder"):
```
katalinic-gc, line 416: Is that the case yet?

katalinic-gc (approved):
lgtm

PR #275, opened 2023-03-03
Layers per ipu improvements

rahult-graphcore:
Changes 
- `layers_per_ipu` to support wildcards (-1). Default value changed from [1] to [-1]. With this change, the `get_layer_ipu` method will use `ipus_per_replica` to fill in the `layers_per_ipu` if the wildcard is used. 
- If `ipus_per_replica` is 1, `layers_per_ipu` is ignored, and all layers are placed on IPU0. If `ipus_per_replica` > 1 and `layers_per_ipu` uses default value [-1] (or [-1] * `ipus_per_replica`), then the layers are evenly split over `ipus_per_replica` IPUs. 
- `layers_per_ipu` can support a combination of integers and wildcards e.g `[1, 1, -1, -1] will put 1 layer each on IPU0 and IPU1, and split the remaining layers evenly between IPU2 and IPU3. If there are an odd number of layers, the extra layer is placed on the last wildcard IPU. 
- Updated the docstring in the IPUConfig class

Tests
- Manually tested on the named-entity-extraction notebook, using different ipu_configs with different settings for `layers_per_ipu` and `ipus_per_replica`, to ensure the result is as expected. 
- Ran test_examples.py overnight, ensured all 30 tests passed.

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -229,8 +229,14 @@ def get_layer_ipu(ipu_config: IPUConfig, target_number_of_layers: Optional[Union
     layers_per_ipu = ipu_config.layers_per_ipu
     ipus_per_replica = ipu_config.ipus_per_replica
 
+    # Check inputs are valid
+    if not all(isinstance(n, int) and n >= -1 for n in layers_per_ipu):
+        raise ValueError("Invalid values in layers_per_ipu. " f"layers_per_ipu={layers_per_ipu}")
+    if ipus_per_replica < 1:
+        raise ValueError("Invalid value for ipus_per_replica. " f"ipus_per_replica={ipus_per_replica}")
+
     if target_number_of_layers is not None:
-        if not isinstance(target_number_of_layers, (int, float)):
+        if not isinstance(target_number_of_layers, int):
             target_number_of_layers = len(target_number_of_layers)
 
         # if ipus_per_replica is 1, then put everything on IPU0, ignoring layers_per_ipu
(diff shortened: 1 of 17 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -225,21 +225,60 @@ def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool
             return sum(p.numel() for p in self.parameters() if p.requires_grad or not only_trainable)
 
 
-def get_layer_ipu(layers_per_ipu: List[int], target_number_of_layers: Optional[Union[int, List]] = None):
+def get_layer_ipu(ipu_config: IPUConfig, target_number_of_layers: Optional[Union[int, List]] = None):
+    layers_per_ipu = ipu_config.layers_per_ipu
+    ipus_per_replica = ipu_config.ipus_per_replica
+
+    if target_number_of_layers is not None:
+        if not isinstance(target_number_of_layers, (int, float)):
```
jimypbr, line 233: why checking for float here?

optimum/graphcore/modeling_utils.py
```diff
@@ -225,21 +225,60 @@ def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool
...
+            target_number_of_layers = len(target_number_of_layers)
+
+        # if ipus_per_replica is 1, then put everything on IPU0, ignoring layers_per_ipu
+        if ipus_per_replica == 1:
+            return [0] * target_number_of_layers
+
+        elif ipus_per_replica > 1:
+            # default/wildcards - split layers evenly over all ipus
+            if layers_per_ipu in ([-1], [-1] * ipus_per_replica):
```
jimypbr, line 242: nice. this handles us just passing `ipus_per_replica=2` and it spreading everything over two IPUs evenly.

jimypbr (changes requested):
Add unit tests for `get_layer_ipu` to check all the edge cases

PR #277, opened 2023-03-07
Add PopART option that enables the use of models with weights exceeding ~2GB

HMellor:
# What does this PR do?

If you try and load a very large model (in my case it was T5-XL), then you run into the following error from protobuf
```
onnx.ModelProto exceeded maximum protobuf size of 2GB
```
This PR adds `opts._Popart.set("saveInitializersToFile", NamedTemporaryFile().name)` which allows this error to be circumvented.

## Before submitting


jimypbr (approved):
LGTM. Running with test-examples hasn't shown any side-effect of having this on all the time.

PR #278, opened 2023-03-08
PackedBERT custom batched inference pipelines to enable inference + usability fixes

arsalanu:
# What does this PR do?

Adds custom pipelines into packed bert notebooks (in the prototype packing util folder). Packing only really benefits batched inference on larger workloads so its effectively a wrapper for a custom inference call (raw text -> predictions, mimicking the behaviour of HF pipelines, but with batching)

Also adds some fixes to retain order of samples for inference (after some realisations when seeing it used in practice, I realised the original implementation couldn't realistically be used for any inference tasks) and changes to the notebooks to expose the batched inference. It also lets the notebooks demonstrate the actual throughput value obtained from packing.


notebooks/packed_bert/pipeline/packed_bert.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def __init__(
+        self,
+        model,
+        executable_cache_dir: str = "./exe_cache",
+        problem_type: str = "question_answering",
+        max_seq_per_pack: int = 12,
+        max_seq_length: int = 384,
+        pretrained_tokenizer: str = "bert-base-uncased",
+        ipu_device_iterations: int = None,
```
jimypbr, line 263: all these `ipu_` arguments. Why don't you just pass an `IPUConfig`?

PR #279, opened 2023-03-09
Speed up examples CI

HMellor:
# What does this PR do?

- Prevents the `venv`s  created for each example from clashing
- Uses a `FileLock` to stop concurrent `pip install`s of `optimum-graphcore`
- Add `POPTORCH_WAIT_FOR_IPU=1` to `test-examples.yml`, this allows:
  - Multiple examples to compile concurrently
  - Maximum IPU utilisation because the IPUs won't be idle while waiting for the next example to compile
- Increase the number of workers in CI to 4


regisss:
This PR should fix the doc build: https://github.com/huggingface/optimum-graphcore/pull/280

HMellor:
Great, thanks for the fix. Will I be ok to land this without the docs build passing? (assuming an approving review)

Also, re my most recent commit, I've reduced the worker count from 16 down to 4 because we likely don't have the RAM/swap to compile 16 models at once.

PR #280, opened 2023-03-09
Fix doc build

regisss:
# What does this PR do?

This PR fixes the doc build which is currently broken because:

- The Docker image does not install the latest node.js version
- The name of your clone repo may not be `optimum-graphcore`, for example here: https://github.com/graphcore/optimum-graphcore-fork

## Before submitting

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -6,7 +6,7 @@ ARG commit_sha
 ARG clone_url
 ARG clone_name
 
-# Need node to build doc HTML. Taken from https://askubuntu.com/a/1113339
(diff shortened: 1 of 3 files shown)


docs/Dockerfile
```diff
@@ -1,19 +1,22 @@
 # Python 3.8 is needed because graphcore/pytorch:latest may be based on Ubuntu 18.04 with Python 3.6
 # And the constraints transformers==4.20.0 is not compatible with Python 3.6
```
jimypbr, line 2: FYI these comments are no longer true. We only support 3.8. Main branch has been upgraded to `transformers==4.25.1`.

PR #281, opened 2023-03-10
Fix PR doc build when the PR comes from a clone with a different name

regisss:
# What does this PR do?

The PR doc build currently fails when the PR comes from a clone with a name which is different from `optimum-graphcore`. This PR fixes that.

## Before submitting


jimypbr (approved):
lgtm

PR #282, opened 2023-03-13
Infer `ipus_per_replica` from `layers_per_ipu`

HMellor:
# What does this PR do?

If the user were to parallelise a model themselves we make them provide `layers_per_ipu` and `ipus_per_replica`. However, `ipus_per_replica` is always just `len(layers_per_ipu)`. So I have made that the default behaviour. You can still provide it explicitly to override this if you need to.

## Before submitting


jimypbr (approved):
LGTM

PR #283, opened 2023-03-14
Run encoder on IPU for encoder-decoder text-gen models

jimypbr:
# What does this PR do?

Optimisation for encoder/decoder text-gen models that runs the encoder part of the model on IPU instead of CPU.
It does this by automatically splitting the IPUConfig into two different configs for the encoder and decoder based upon the `ipu_config.layers_per_ipu` and the corresponding `num_encoder_layers` and `num_decoder_layers` from the model.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/generation_utils.py
+++ b/optimum/graphcore/generation_utils.py
@@ -12,8 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import os
 import copy
+import json
 import warnings
+import contextlib
 from itertools import groupby
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
@@ -48,6 +51,19 @@ from .modeling_utils import get_layer_ipu
 
 
 logger = logging.get_logger(__name__)
+
+
+@contextlib.contextmanager
+def graph_profile_dir_append(append: str):
+    if poplar_engine_options := os.getenv("POPLAR_ENGINE_OPTIONS"):
+        poplar_engine_options_dict = json.loads(poplar_engine_options)
+        poplar_engine_options_dict["autoReport.directory"] += append
+        os.environ["POPLAR_ENGINE_OPTIONS"] = json.dumps(poplar_engine_options_dict)
+    try:
+        yield
+    finally:
+        if poplar_engine_options:
+            os.environ["POPLAR_ENGINE_OPTIONS"] = poplar_engine_options
 
 
 def _split_encoder_decoder_ipu_config(ipu_config, config):
@@ -132,7 +148,8 @@ class IPUGenerationMixin(GenerationMixin):
                 )
 
         # This will trigger a compile first time it's ran
-        return self.poptorch_decoder(*args, **kwargs)
+        with graph_profile_dir_append("/decoder"):
+            return self.poptorch_decoder(*args, **kwargs)
 
     def _prepare_encoder_decoder_kwargs_for_generation(
         self, inputs_tensor: torch.Tensor, model_kwargs, model_input_name: Optional[str] = None
@@ -167,7 +184,8 @@ class IPUGenerationMixin(GenerationMixin):
                 encoder.eval(), self.encoder_ipu_config.to_options(for_inference=True)
             )
             self.ipu_config = self.ipu_config
-        model_kwargs["encoder_outputs"]: ModelOutput = self.poptorch_encoder(**encoder_kwargs)
+        with graph_profile_dir_append("/encoder"):
+            model_kwargs["encoder_outputs"]: ModelOutput = self.poptorch_encoder(**encoder_kwargs)
 
         return model_kwargs
 
--- a/optimum/graphcore/generation_utils.py
+++ b/optimum/graphcore/generation_utils.py
@@ -140,12 +140,7 @@ class IPUGenerationMixin(GenerationMixin):
         # 1. get encoder
         encoder = self.get_encoder()
 
-        # 2. get the model dtype and cast encoder to fp32 for cpu
-        dtype = next(encoder.parameters()).dtype
-        if dtype is torch.float16:
-            encoder.to(torch.float32)
-
-        # 3. prepare encoder args and encoder kwargs from model kwargs
+        # 2. prepare encoder args and encoder kwargs from model kwargs
         irrelevant_prefix = ["decoder_", "cross_attn", "use_cache"]
         encoder_kwargs = {
             argument: value
@@ -153,13 +148,13 @@ class IPUGenerationMixin(GenerationMixin):
             if not any(argument.startswith(p) for p in irrelevant_prefix)
         }
 
-        # 4. Split self.ipu_config into encoder and decoder ipu_configs
+        # 3. Split self.ipu_config into encoder and decoder ipu_configs
         if not hasattr(self, "encoder_ipu_config"):
             self.encoder_ipu_config, self.decoder_ipu_config = _split_encoder_decoder_ipu_config(
                 self.ipu_config, self.config
             )
 
-        # 5. make sure that encoder returns `ModelOutput`
+        # 4. make sure that encoder returns `ModelOutput`
         model_input_name = model_input_name if model_input_name is not None else self.main_input_name
         encoder_kwargs["return_dict"] = True
         encoder_kwargs[model_input_name] = inputs_tensor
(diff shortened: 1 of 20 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -225,6 +225,32 @@ def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool
...
+        "'{arg}' must take one of the following three forms:\n"
+        "  - float: one value for all parts\n"
+        "  - list(float): different values for each parts\n"
+        "  - list(list(float)): different values for each IPU in each part\n"
+        "The outer list must be of length {n}"
+    )
+    # Split layers_per_ipu between the given parts
+    ipu_configs = {part: copy.deepcopy(ipu_config) for part in parts}
+    for i, (name, ipu_config) in enumerate(ipu_configs.items()):
+        if isinstance(ipu_config.layers_per_ipu, list):
```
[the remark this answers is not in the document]
HMellor: If the user passes an integer, do we put it into a list?

optimum/graphcore/modeling_utils.py
```diff
@@ -225,6 +225,32 @@ def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool
             return sum(p.numel() for p in self.parameters() if p.requires_grad or not only_trainable)
 
 
+def split_ipu_config(ipu_config: IPUConfig, parts: List[str] = []) -> List[IPUConfig]:
+    msg = (
```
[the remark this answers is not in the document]
HMellor: Is this comment in the right place? I'm not sure I follow

tests/test_ipu_configuration.py
```diff
@@ -304,3 +304,53 @@ def test_layers_per_ipu(self):
         ipu_config = IPUConfig(ipus_per_replica=0)
         with pytest.raises(ValueError, match=r"Invalid value for ipus_per_replica"):
             layer_ipu = get_layer_ipu(ipu_config, 6)
+
+    def test_split_encoder_decoder_ipu_config(self):
```
HMellor, line 308: All the tests are good, less keen that they are all in one monolithic test.

Splitting them up and [grouping them in a class](https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class) would be good.

optimum/graphcore/generation_utils.py
```diff
@@ -80,45 +100,49 @@ def _pad_tensors_to_max_len(self, tensor: torch.Tensor, max_length: int, pad_tok
...
         encoder_kwargs[model_input_name] = inputs_tensor
-        model_kwargs["encoder_outputs"]: ModelOutput = encoder(**encoder_kwargs)
-
-        # 5. undo the casting for cpu
-        if dtype is torch.float16:
-            encoder.to(torch.float16)
-            for key, output in model_kwargs["encoder_outputs"].items():
-                if isinstance(output, torch.Tensor) and output.dtype is torch.float32:
-                    model_kwargs["encoder_outputs"][key] = output.to(torch.float16)
+        if not hasattr(self, "poptorch_encoder"):
```
HMellor, line 130: Can you either use this in both places or use `x_ipu_config = getattr(self, "x_ipu_config", self.ipu_config)` in both places like you did for the decoder?

optimum/graphcore/generation_utils.py
```diff
@@ -12,7 +12,12 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import contextlib
+import copy
```
HMellor, line 16: Doesn't look like this is used based on the diff

optimum/graphcore/generation_utils.py
```diff
@@ -12,7 +12,12 @@
...
+import json
+import os
 import warnings
+from itertools import groupby
```
HMellor, line 20: No longer used

optimum/graphcore/generation_utils.py
```diff
@@ -80,45 +100,49 @@ def _pad_tensors_to_max_len(self, tensor: torch.Tensor, max_length: int, pad_tok
...
+            # Use split encoder ipu_config for encoder/decoder models
+            self.poptorch_encoder = poptorch.inferenceModel(
+                encoder.eval(), self.encoder_ipu_config.to_options(for_inference=True)
+            )
+        with graph_profile_dir_append("/encoder"):
+            model_kwargs["encoder_outputs"]: ModelOutput = self.poptorch_encoder(**encoder_kwargs)
 
         return model_kwargs
 
+    def detachFromDevice(self):
```
HMellor, line 140: Will this cause problems if a pipeline is created with a non split model? Or is this class only used for split models?

optimum/graphcore/modeling_utils.py
```diff
@@ -241,7 +253,7 @@ def get_layer_ipu(ipu_config: IPUConfig, target_number_of_layers: Optional[Union
 
         # if ipus_per_replica is 1, then put everything on IPU0, ignoring layers_per_ipu
         if ipus_per_replica == 1:
-            return [0] * target_number_of_layers
+            return [target_number_of_layers]
```
HMellor, line 256: Good catch

optimum/graphcore/modeling_utils.py
```diff
@@ -278,6 +290,76 @@ def get_layer_ipu(ipu_config: IPUConfig, target_number_of_layers: Optional[Union
...
+    """
+    # Need at least two IPUs to do the split
+    if ipu_config.ipus_per_replica < 2:
+        raise ValueError("Need ipus_per_replica of at least 2 to split ipu_config into encoder and decoder configs")
+
+    ipu_configs = {name: copy.deepcopy(ipu_config) for name in ["encoder", "decoder"]}
+
+    # Split layers_per_ipu between the given num layers
+    layers_per_ipu = _expand_layers_per_ipu_wildcard(ipu_config, num_encoder_layers + num_decoder_layers)
+    cumsum = [sum(layers_per_ipu[: i + 1]) for i in range(len(layers_per_ipu))]
```
HMellor, line 334: What happens if there's a `0` at the cut?

i.e. if we had `[1, 2, 3, 0, 5, 6, 7, 8]` would it get split into `[1, 2, 3]` and `[0, 5, 6, 7, 8]`?

HMellor (changes requested):

optimum/graphcore/generation_utils.py
jimypbr: The decoder one is a special case that needs that check because models like GPT are decoder-only

optimum/graphcore/generation_utils.py
jimypbr: This might cause problems if the model is being used like in training mode in the `IPUTrainer`.

optimum/graphcore/modeling_utils.py
jimypbr: Good catch!

tests/test_ipu_configuration.py
HMellor: I think I'll do this in a separate PR to make this change on other tests too (if I feel it's worth it)

optimum/graphcore/generation_utils.py
HMellor: I suppose if the model isn't split it'll just be a no-op, but we'd like it to work for all possible poplar executors that can exist in generation

optimum/graphcore/pipelines/__init__.py
```diff
@@ -226,7 +226,10 @@ def get_poplar_executor(
     ipu_config.inference_replication_factor = 1
     try:
         model = to_pipelined(model, ipu_config, force=False)
-        model.parallelize()
+        if model.config.is_encoder_decoder:
+            model.parallelize(for_generation=True)
+        else:
+            model.parallelize()
```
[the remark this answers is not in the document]
jimypbr: No the `is_encoder_decoder` is a taxonomy of the model and doesn't change at run time. We need a `for_generation` so we can run encoder-decoder models in either "training mode" (encoder and decoder in same executor) or in generation mode.

optimum/graphcore/generation_utils.py
```diff
@@ -42,10 +45,25 @@
 from transformers.modeling_outputs import ModelOutput
 from transformers.pytorch_utils import torch_int_div
 
+from .modeling_utils import get_layer_ipu
```
HMellor, line 48: Not used here?
```suggestion
```

optimum/graphcore/generation_utils.py
HMellor: Did you check this?

optimum/graphcore/modeling_utils.py
```diff
@@ -225,23 +231,35 @@ def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool
...
+    ```
+    this function will expand the wildcard values to `layers_per_ipu = [4, 5]`
+    """
+    layers_per_ipu = copy.deepcopy(ipu_config.layers_per_ipu)
     ipus_per_replica = ipu_config.ipus_per_replica
 
     # Check inputs are valid
     if not all(isinstance(n, int) and n >= -1 for n in layers_per_ipu):
-        raise ValueError("Invalid values in layers_per_ipu. " f"layers_per_ipu={layers_per_ipu}")
+        raise IncompatibleIPUConfigError("Invalid values in layers_per_ipu. " f"layers_per_ipu={layers_per_ipu}")
```
HMellor, line 252:
```suggestion
        raise IncompatibleIPUConfigError(f"Invalid values in layers_per_ipu. layers_per_ipu={layers_per_ipu}")
```

PR #285, opened 2023-03-21
Use buffers to cache the encoder hidden states in decoder wrapper

jimypbr:
# What does this PR do?

POC using buffers to store the encoder hidden states in the decoder wrapper.
The current implementation copies all the weights and buffers every time the encoder is ran, so isn't efficient yet.

Fixes # (issue)

## Before submitting


katalinic-gc (approved):
lgtm

PR #287, opened 2023-03-21
Fix encoder-decoder splitting for cases where ipus have 0 layers

jimypbr:
# What does this PR do?

Fixes # (issue)

Now correctly splits cases where `layers_per_ipu` has 0s in it. Examples:
```
ipu_config = IPUConfig(layers_per_ipu=[1, 2, 3, 0, 5, 6, 7, 8]) -> [1, 2, 3, 0] / [5, 6, 7, 8]
ipu_config = IPUConfig(layers_per_ipu=[0, 6, 0, 6]) -> [0, 6] / [0, 6]
ipu_config = IPUConfig(layers_per_ipu=[3, 0, 3, 0, 0, 7]) -> [3, 0, 3, 0] / [0, 7]
```

## Before submitting

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -350,6 +350,9 @@ def split_encoder_decoder_ipu_config(
             f"\tipu_config.layers_per_ipu={ipu_config.layers_per_ipu}\n"
             f"\tnum_encoder_layers={num_encoder_layers}\n"
             f"\tnum_decoder_layers={num_decoder_layers}\n"
+            "Possible causes: \n"
+            "Encoder and decoder layers cannot be placed on the same IPUs.\n"
+            "The encoder and decoder layers_per_ipu splits each need a number of devices that's a power of 2."
         )
     ipu_configs["encoder"].layers_per_ipu = layers_per_ipu[:cut]
     ipu_configs["decoder"].layers_per_ipu = layers_per_ipu[cut:]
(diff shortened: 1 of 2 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -341,7 +341,8 @@ def split_encoder_decoder_ipu_config(
     layers_per_ipu = _expand_layers_per_ipu_wildcard(ipu_config, num_encoder_layers + num_decoder_layers)
     cumsum = [sum(layers_per_ipu[: i + 1]) for i in range(len(layers_per_ipu))]
     try:
-        cut = [i + 1 for i, c in enumerate(cumsum) if c == num_encoder_layers][-1]
+        cut = [i + 1 for i, c in enumerate(cumsum) if c == num_encoder_layers]
+        cut = max([num for num in cut if num & (num - 1) == 0])
     except:
         raise IncompatibleIPUConfigError(
             f"Unable to find valid split of ipu_config.layers_per_ipu\n"
```
kundaMwiza, line 348: Cool. Something minor, can we update the error to let the user know that:
- for generation, they cannot place enocder and decoder layers on the same IPU
- the last encoder layer has to occur on an IPU that is a power of 2

jimypbr: > for generation, they cannot place enocder and decoder layers on the same IPU
Yep I'll add that to the error message

> the last encoder layer has to occur on an IPU that is a power of 2
I don't think that's the case. If, for some reason, you have the edge case of `layers_per_ipu=[1, 2, 3, 0, 0, 2]` it will cut it at the next power of 2. So that will become `[1, 2, 3, 0] / [0, 2]`

kundaMwiza: yep, I guess I meant something that tells the user about the fact that the encoder and decoder both need ipu devices that have 2**n ipus

kundaMwiza (approved):
LGTM

PR #293, opened 2023-03-22
Pin the optimum version requirement

jimypbr:
## Before submitting

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ First clone and install `optimum`:
 ```
 git clone https://github.com/huggingface/optimum
 cd optimum
-git checkout v1.7.1-release
+git checkout v1.6.2
 pip install -e .
 cd ..
 ```
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ If you are working on the `optimum-graphcore` code then you should use an editab
 First clone and install `optimum`:
 
 ```
-git clone https://github.com/huggingface/optimum --branch v1.6.2-release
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,6 +35,7 @@ First clone and install `optimum`:
 ```
 git clone https://github.com/huggingface/optimum
 cd optimum
+git checkout v1.7.1-release
```
HMellor, line 38: `--branch` can take tags too? (head will be detached, but not a problem since we only do this so that we can install optimum-graphcore editable)
```suggestion
git clone https://github.com/huggingface/optimum --branch v1.7.1-release
cd optimum
```

setup.py
```diff
@@ -14,10 +14,10 @@
 
 INSTALL_REQUIRES = [
     "transformers==4.25.1",
-    "optimum",
+    "optimum==1.6.1",
     "datasets",
     "tokenizers",
-    "torch",
+    "torch @ https://download.pytorch.org/whl/cpu/torch-1.13.0%2Bcpu-cp38-cp38-linux_x86_64.whl",
```
HMellor, line 20:
```suggestion
    "torch @ https://download.pytorch.org/whl/cpu/torch-1.13.1%2Bcpu-cp38-cp38-linux_x86_64.whl",
```

jimypbr: For SDK3.1 it's 1.13.0. I'll change it on the sdk3.2 branch to 1.13.1.

HMellor: Ok, will you make a patch release for SDK3.1 with this change?

PR #295, opened 2023-03-23
Text Gen slice decoder projection optimisation

jimypbr:
# What does this PR do?

Optimization of the decoder-only forward mode used in text generation.
Only computes the projection on the position `t` of the sequence rather than on the whole sequence.
This saves a lot of memory in the LM-head matmul.

Fixes # (issue)

## Before submitting


optimum/graphcore/generation_utils.py
```diff
@@ -81,12 +102,15 @@ def forward(self, t, **model_inputs):
         Returns:
             The output logits at position `t` only
         """
+        # Update the cur_step buffer in the _SliceLinear layer
+        self.pipelined_model.get_output_embeddings().cur_step.copy_(t)
```
katalinic-gc, line 106: I am still contemplating if there's a cleaner way to do this

optimum/graphcore/generation_utils.py
```diff
@@ -63,6 +63,25 @@ def graph_profile_dir_append(append: str):
...
+    `self.cur_step` is a buffer.
+    """
+
+    def __init__(self, linear_layer):
+        super().__init__()
+        self.wrapped_linear = linear_layer
+        self.register_buffer("cur_step", torch.tensor([0], dtype=torch.int32))
+
+    def forward(self, x):
+        x = poptorch.dynamic_slice(x, 1, self.cur_step, 1, 1)
```
katalinic-gc, line 81: Just wondering, will seq length always be in axis 1? Do we ensure in beam search e.g. that beams go into batch dim?

optimum/graphcore/generation_utils.py
```diff
@@ -63,6 +63,25 @@ def graph_profile_dir_append(append: str):
...
+    before executing the linear. The intended use is as an optimized replacement of the
+    LM Head in the Decoder for text generation inference.
+    The slice is performed on the position `self.cur_step` of the input tensor.
+    `self.cur_step` is a buffer.
+    """
+
+    def __init__(self, linear_layer):
+        super().__init__()
+        self.wrapped_linear = linear_layer
+        self.register_buffer("cur_step", torch.tensor([0], dtype=torch.int32))
```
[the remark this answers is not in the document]
jimypbr: yeah I've tested this with gpt2, bart, and t5 and they work as before.

PR #304, opened 2023-03-31
SerializedEmbedding: override default freeze=True on deserialization

kundaMwiza:
# What does this PR do?

The default behaviour of `nn.Embedding.from_pretrained` is to freeze the constructed embedding weights. This may not be what the user intends to do.  This PR overrides the default value and allows the user to optionally change the behaviour (for example in `pipelined_model_cls.deparallelize()`)

## Before submitting

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -421,15 +421,18 @@ class SerializedEmbedding(nn.Module):
             ]
         )
 
-    def deserialize(self):
+    def deserialize(self, freeze=False):
         """
         Deserialize the internal wrapped embedding layer and return it as a
-        `nn.Embedding` object.
+        `nn.Embedding` object. Optionally override the default behaviour of
+        freezing embeddings when initialising from a tensor object.
 
         Returns:
             `nn.Embedding` layer
         """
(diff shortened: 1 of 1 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -421,15 +421,18 @@ def __init__(self, embedding: nn.Embedding, serialization_factor: int):
...
-        `nn.Embedding` object.
+        `nn.Embedding` object. Optionally override the default behaviour of
+        freezing embeddings when initialising from a tensor object.
 
         Returns:
             `nn.Embedding` layer
         """
-        return nn.Embedding.from_pretrained(torch.vstack([l.weight for l in self.split_embeddings]), padding_idx=0)
+        return nn.Embedding.from_pretrained(
+            torch.vstack([l.weight for l in self.split_embeddings]), padding_idx=0, freeze=freeze
```
jimypbr, line 434: Could we inspect the `weight.requires_grad` to determine whether to freeze? Same with the `serialize()` function.

kundaMwiza: Makes more sense, thanks

jimypbr (approved):

PR #308, opened 2023-04-06
Enable training and inference specific configurations using a single `IPUConfig`

HMellor:
# What does this PR do?

Currently the PR creates `inference_` and `training_` versions of: 
- `layers_per_ipu`
- `ipus_per_replica`
- `matmul_proportion`

in the `IPUConfig` constructor, which are populated either:
- From being explicitly defined in the constructor’s `kwargs`
- By defaulting to the non `inference_` or `training_` (this makes the PR backwards compatible with configs that don’t specify the execution mode)

Which version gets returned is decided by managed attributes created using Python descriptors ([Descriptor HowTo Guide](https://docs.python.org/3/howto/descriptor.html#managed-attributes)). i.e. `ipu_config.layers_per_ipu` will return different things depending on if `ipu_config.mode` is `training` or `inference`. This means that as long as the correct `mode` is set before `model.parallelize()` is called, the attributes for the correct mode will be retrieved by the descriptor’s getter.

This mode selection is currently done in the `pipeline` constructor and right before the `training_model` and `inference_model` is created in the `IPUTrainer`

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -133,10 +133,12 @@ class IPUConfig(BaseConfig):
 
         def __set__(self, obj, value):
             if isinstance(obj, IPUConfig):
+                logger.info(f"ManagedAttribute {self.attr} writing to {obj.mode}_{self.attr}")
                 return setattr(obj, f"{obj.mode}_{self.attr}", value)
 
         def __get__(self, obj, objtype=None):
             if isinstance(obj, IPUConfig):
+                logger.info(f"ManagedAttribute {self.attr} reading from {obj.mode}_{self.attr}")
                 return getattr(obj, f"{obj.mode}_{self.attr}")
 
     # Create descriptor based managed attributes which will either return the
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -166,8 +166,7 @@ class IPUConfig(BaseConfig):
 
             # If matmul_proportion is a list and its length is not equal to {mode}_ipus_per_replica, use the
             # default float value for default_matmul_prop instead
-            if (isinstance(matmul_prop, list) and
-                len(matmul_prop) != getattr(self, f"{mode}_ipus_per_replica")):
+            if isinstance(matmul_prop, list) and len(matmul_prop) != getattr(self, f"{mode}_ipus_per_replica"):
                 matmul_prop = 0.2
             setattr(self, f"{mode}_matmul_proportion", kwargs.pop(f"{mode}_matmul_proportion", matmul_prop))
 
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -132,14 +132,16 @@ class IPUConfig(BaseConfig):
             self.attr = attr
 
         def __set__(self, obj, value):
-            logger.info(f"ManagedAttribute {self.attr} writing to {obj.mode}_{self.attr}")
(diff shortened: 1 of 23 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -304,8 +304,8 @@ def split_encoder_decoder_ipu_config(
 
     ipu_configs = {name: copy.deepcopy(ipu_config) for name in ["encoder", "decoder"]}
 
-    # Split layers_per_ipu between the given num layers
-    layers_per_ipu = _expand_layers_per_ipu_wildcard(ipu_config, num_encoder_layers + num_decoder_layers)
+    # Split inference_layers_per_ipu between the given num layers (this function is only called for generation)
+    layers_per_ipu = _expand_layers_per_ipu_wildcard(ipu_config, num_encoder_layers + num_decoder_layers, True)
```
HMellor, line 308:
```suggestion
    layers_per_ipu = _expand_layers_per_ipu_wildcard(ipu_config, num_encoder_layers + num_decoder_layers, for_training=False)
```

HMellor:
Not sure why the `mode` property isn't appearing in the docs, they should be supported https://github.com/huggingface/doc-builder/pull/87

optimum/graphcore/ipu_configuration.py
```diff
@@ -119,11 +127,49 @@ class IPUConfig(BaseConfig):
...
-        self.layers_per_ipu = kwargs.pop("layers_per_ipu", [-1])
-        self.ipus_per_replica = kwargs.pop("ipus_per_replica", len(self.layers_per_ipu))
+        # Get execution mode agnostic arguments
+        layers_per_ipu = kwargs.pop("layers_per_ipu", [-1])
+        ipus_per_replica = kwargs.pop("ipus_per_replica", len(layers_per_ipu))
+        matmul_proportion = kwargs.pop("matmul_proportion", 0.2)
+
+        # Get execution mode specific arguments (if available)
+        for mode in self.modes:
+            setattr(self, f"{mode}_layers_per_ipu", kwargs.pop(f"{mode}_layers_per_ipu", layers_per_ipu))
```
katalinic-gc, line 160: nitpicking, would you mind inserting a space after this and the next `setattr` to visually separate the three attributes

HMellor:
> do we need to establish what is more common to determine what the default should be? 

Currently the default `mode` is training because:
- The default mode for `torch.nn.Module` is also training
-  IPU pipelines made for training will always work for inference, but not the other way around
Therefore, the performance will either be the same or better than it is now:
- If you are using the `IPUTrainer` or `Pipeline`
- Depending on the advanced developers knowledge of the IPUConfig

> what the most appropriate place would be
Yeah, unfortunately `to_options()` is too late for the `mode` to be set because it needs to be set to the correct value before `parallelize()` is called

> maybe the logic in `IPUConfig.to_options` should also be adapted eventually as it currently does the `for_inference` checks explicitly.

Yeah I agree with this. We would need to change the `if for_inference` to `if self.mode == "inference"` and turn `replication_factor` and `device_iterations` into `ManagedAttributes`

optimum/graphcore/ipu_configuration.py
HMellor: Yeah that's fine by me, more readable is more better

katalinic-gc (approved):

jimypbr:
Could you app `nn.Module` and also have `IPUConfig.train()` and `IPUConfig.eval()`?

HMellor:
> Could you ape `nn.Module` and have `ipu_config.train()` and `ipu_config.eval()` instead of `ipu_config.mode = "train"`

Good suggestion, done

PR #310, opened 2023-04-06
Generation projection layer fixes

kundaMwiza:
# What does this PR do?

- Do not add index buffer to the model state dict when wrapping the projection layer with `_IndexedInputLinear`
- Ensure projection layer is always wrapped for calls to the decoder

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/generation_utils.py
+++ b/optimum/graphcore/generation_utils.py
@@ -75,7 +75,7 @@ class _IndexedInputLinear(nn.Module):
     def __init__(self, linear_layer):
         super().__init__()
         self.wrapped_linear = linear_layer
-        self.register_buffer("index", torch.tensor([0], dtype=torch.int32))
+        self.register_buffer("index", torch.tensor([0], dtype=torch.int32), persistent=False)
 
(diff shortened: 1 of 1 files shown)


optimum/graphcore/generation_utils.py
```diff
@@ -119,7 +119,14 @@ def _pad_tensors_to_max_len(self, tensor: torch.Tensor, max_length: int, pad_tok
 
     def _call_generate(self, *args, cur_token_id: int, **kwargs):
         t = self._get_cur_token_logits_tensor(cur_token_id)
-        if not hasattr(self, "poptorch_decoder"):
+
+        if hasattr(self, "poptorch_decoder"):
+            # make sure that the projection layer is still wrapped. it can be unwrapped
+            # on model deparallelization calls
```
katalinic-gc, line 125: Can you describe the flow in which this would occur?

katalinic-gc: https://github.com/huggingface/optimum-graphcore/pull/319

PR #352, opened 2023-04-20
Minor changes to make whisper-example.ipynb consistent with template updates

jayniep-gc:
There have been some updates to the notebooks template. This PR implements those updates in this notebook.


evawGraphcore (approved):
LGTM

PR #353, opened 2023-04-20
Add pipeline mode attribute

kundaMwiza:
# What does this PR do?

- Avoids redundant deparallelize / parallelize
- On the first instance of the call `trainer.train()` the model has already been parallelized in the constructor of the trainer. when `deparallelize` & `parallelize` are called in `wrap_model`, the models parameters have already been provided to the optimizer.  Parameter changes in `deparallelize` / `parallelize` such as restoring the `lm_head` with a newly constructed `torch.nn.Linear` can be disastrous, since the address in memory of the new weights is not necessarily the same as the weights previously provided to the optimizer. Differences in memory address of weights will lead to errors when `poptorch` attempts to compile the model. For example, currently any model that does not have `tied weights` and uses `SerializedEmbedding` or `SerializedLinear` will fail.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -191,7 +191,7 @@ class PipelineMixin:
             if m is not self:
                 poptorch.removeBlocks(m)
         # Model is no longer parallelized
-        self.pipeline_mode = "default"
+        self.pipeline_mode = "deparallelized"
         return self
 
     def num_parameters(self, only_trainable: bool = False, exclude_embeddings: bool = False) -> int:
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -78,8 +78,48 @@ def to_pipelined(model: nn.Module, ipu_config: IPUConfig, force: bool = False):
         else:
             raise KeyError(f"{model_cls.__name__} pipelined version not found in registry.")
 
-
 class PipelineMixin:
+    ALLOWED_PIPELINE_MODES = ("train", "generation", "evaluation", "default")
+    _pipeline_mode = "default"
+    
+    @property
+    def pipeline_mode(self):
+        return self._pipeline_mode
+    
+    @pipeline_mode.setter
+    def pipeline_mode(self, new_mode):
+        assert new_mode in self.ALLOWED_PIPELINE_MODES, f"`new_mode` can only be one of {self.ALLOWED_PIPELINE_MODES=}. You provided {new_mode}."
+        self._pipeline_mode = new_mode 
+        return self
+    
+    def _set_pipeline_mode_train(self):
+        self.pipeline_mode = "train"
+        return self
+
+    def _set_pipeline_mode_generation(self):
+        self.pipeline_mode = "generation"
+        return self
+    
+    def _set_pipeline_mode_evaluation(self):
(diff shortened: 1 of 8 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -80,6 +80,21 @@ def to_pipelined(model: nn.Module, ipu_config: IPUConfig, force: bool = False):
 
 
 class PipelineMixin:
+    ALLOWED_PIPELINE_MODES = ("train", "generation", "evaluation", "deparallelized")
```
HMellor, line 83: Why all caps? That usually signifies a global variable, but this isn't one of those

optimum/graphcore/modeling_utils.py
```diff
@@ -80,6 +80,21 @@ def to_pipelined(model: nn.Module, ipu_config: IPUConfig, force: bool = False):
...
+    _pipeline_mode = "deparallelized"
```
HMellor, line 84: Should this be an instance variable instead of a class variable?

optimum/graphcore/modeling_utils.py
```diff
@@ -80,6 +80,21 @@ def to_pipelined(model: nn.Module, ipu_config: IPUConfig, force: bool = False):
...
+
+    @property
+    def pipeline_mode(self):
+        return self._pipeline_mode
+
+    @pipeline_mode.setter
+    def pipeline_mode(self, new_mode):
+        assert (
+            new_mode in self.ALLOWED_PIPELINE_MODES
```
HMellor, line 93: Why all caps? That usually signifies a global variable, but this isn't one of those

optimum/graphcore/trainer_seq2seq.py
```diff
@@ -49,7 +50,8 @@ def _wrap_and_compile_model_for_evaluation(self, dataloader, prediction_loss_onl
         return self.model
 
     def _rewrap_model_for_training(self):
-        self.model = self.model.deparallelize().parallelize()
+        if not self.model.pipeline_mode == "train":
+            self.model = self.model.deparallelize().ipu_config.train().parallelize()
```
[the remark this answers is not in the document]
HMellor: Yeah, currently this won't work

optimum/graphcore/modeling_utils.py
```diff
@@ -148,10 +152,11 @@ def ipu_config(self, value: IPUConfig):
             raise TypeError(f"ipu_config must be an instance of IPUConfig, but {type(value)} was provided")
         self._ipu_config = value
 
-    def parallelize(self):
+    def parallelize(self, for_generation=False):
```
HMellor, line 155:
```suggestion
    def parallelize(self):
```

optimum/graphcore/modeling_utils.py
```diff
@@ -434,7 +440,7 @@ def deserialize(self):
 
         freeze = not self.split_embeddings[0].weight.requires_grad
         return nn.Embedding.from_pretrained(
-            torch.vstack([l.weight for l in self.split_embeddings]), padding_idx=0, freeze=freeze
+            torch.vstack([l.weight.detach() for l in self.split_embeddings]), padding_idx=0, freeze=freeze
```
HMellor, line 443: Was this committed by mistake?

PR #355, opened 2023-04-23
Support sharding serialized layers across ipus

kundaMwiza:
# What does this PR do?

- Adding the necessary IPU configuration options that will enable sharding serialized layers like `SerializedEmbedding` and `SplitLinear` across IPUs by using a pipeline configuration similar to `layers_per_ipu`. For example `[0, 0, 1, 2]` as an argument to `serialized_projection_splits_per_ipu` will result in 1 shard of a linear layer being placed on IPU2 and the last 2 shards on the last IPU.  The pipelines will also be split in a similar way for generation. For example if `layers_per_ipu=[0, 2, 2, 0]`, the corresponding encoder and decoder pipelines will be `[0, 2]` and `[2, 0]`. Similarly, `serialized_projection_layers_per_ipu` will be `[0, 0]` and `[1, 2]` for the encoder and decoder respectively. However the encoder portion `[0, 0]` is overrided with `None` as it is no longer a valid value for `serialized_projection_layers_per_ipu`. This behaviour is also similar for `SerializedEmbedding` and `serialized_embedding_splits_per_ipu` 
- Added `projection_serialization_factor` as a new parameter than can be used to serialize the LM head linear layers via `SerializedLinear` or `SplitProjection`. Some architectures rely on the use of `embedding_serialization_factor` for serializing linear layers, however that differs from the documented use case of `embedding_serialization_factor` in the `IPUConfig`. 

This feature will be used in the the `parallelize` method of a model:

```
if ipu_config._serialized_projection_splits_per_ipu:
     serialization_factor = sum(ipu_config._serialized_projection_splits_per_ipu)
     lm_head = SplitProjection(self.lm_head,  serialization_factor)
     lm_head.parallelize(ipu_config._serialized_projection_splits_per_ipu)
```

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -194,7 +194,6 @@ class IPUConfig(BaseConfig):
     _projection_serialization_factor = ManagedAttribute("linear_serialization_factor")
     _serialized_embedding_splits_per_ipu = ManagedAttribute("serialized_embedding_splits_per_ipu")
     _serialized_projection_splits_per_ipu = ManagedAttribute("serialized_linear_splits_per_ipu")
-    
 
     def contents_geq_value_validator(
         name: str, value: Union[float, int, Sequence], floor_value: Union[float, int]
@@ -205,7 +204,10 @@ class IPUConfig(BaseConfig):
         For Union[float, int], ensure the scalar is >= floor_value
         """
 
-        if isinstance(value, Sequence):
+        # Do nothing for optional types
+        if value is None:
+            return
+        elif isinstance(value, Sequence):
             if not all([elem >= floor_value for elem in value]):
                 raise ValueError(
                     f"`IPUConfig` attribute `{name}` must have all elements >= {floor_value}. You provided {value=}"
@@ -231,6 +233,9 @@ class IPUConfig(BaseConfig):
         ("ipus_per_replica", 1),
         ("inference_ipus_per_replica", 1),
         ("embedding_serialization_factor", 1),
+        ("inference_embedding_serialization_factor", 1),
+        ("projection_serialization_factor", 1),
+        ("inference_projection_serialization_factor", 1),
         ("device_iterations", 1),
         ("inference_device_iterations", 1),
     ):
@@ -245,9 +250,36 @@ class IPUConfig(BaseConfig):
             )
 
     attribute_validators["output_mode"].append(output_mode_validator)
-    
+
     def serialized_layer_splits_per_ipu_validator(name: str, value: str):
-        pass
+        """
+        Validates serialized_{projection/embedding}_splits_per_ipu attributes.
+        If `value` is not None. `value` must be of type List[int>=0] with
+        atleast 1 split on 1 IPU. Further splits in the pipeline must be
+        consecutive.
+        """
+
+        if value is None:
+            return
+
+        IPUConfig.contents_geq_value_validator(name, value, floor_value=0)
+
+        # There must be atleast 1 split when the pipeline is provided
+        if sum(value) < 1:
+            raise ValueError(f"`{name}` must have atleast 1 split on 1 IPU.")
+
+        # Check that splits are on consecutive IPUs (e.g. [3,0,2,0] is not allowed)
+        for i, splits in enumerate(value[:-1]):
+            if splits and value[i + 1] == 0 and sum(value[i + 1 :]) == 0:
+                raise ValueError(f"{name}={value} must have its splits on consecutive IPUs")
+
+    for attr in (
+        "serialized_projection_splits_per_ipu",
+        "inference_serialized_projection_splits_per_ipu",
+        "serialized_embedding_splits_per_ipu",
+        "inference_serialized_embedding_splits_per_ipu",
+    ):
+        attribute_validators[attr].append(serialized_layer_splits_per_ipu_validator)
 
     def __init__(
         self,
@@ -339,22 +371,17 @@ class IPUConfig(BaseConfig):
 
         check_and_set_replication_factor("replication_factor", replication_factor)
         check_and_set_replication_factor("inference_replication_factor", inference_replication_factor)
-        
+
+        # Non-transformer layers initialisation
         self.embedding_serialization_factor = embedding_serialization_factor
         self.inference_embedding_serialization_factor = inference_embedding_serialization_factor
         self.serialized_embedding_splits_per_ipu = serialized_embedding_splits_per_ipu
         self.inference_serialized_embedding_splits_per_ipu = inference_serialized_embedding_splits_per_ipu
+
         self.projection_serialization_factor = projection_serialization_factor
         self.inference_projection_serialization_factor = inference_projection_serialization_factor
         self.serialized_projection_splits_per_ipu = serialized_projection_splits_per_ipu
-        self.inference_serialized_projection_splits_per_ipu = inference_serialized_projection_splits_per_ipu    
-
-        self.gradient_accumulation_steps = gradient_accumulation_steps
-        self.device_iterations = device_iterations
-        self.inference_device_iterations = inference_device_iterations
-        self.optimizer_state_offchip = optimizer_state_offchip
-        self.replicated_tensor_sharding = replicated_tensor_sharding
-        self.auto_loss_scaling = auto_loss_scaling
+        self.inference_serialized_projection_splits_per_ipu = inference_serialized_projection_splits_per_ipu
 
         if "sharded_execution_for_inference" in kwargs:
             warnings.warn(
@@ -364,11 +391,18 @@ class IPUConfig(BaseConfig):
         if "enable_half_first_order_momentum" in kwargs:
             warnings.warn('The "enable_half_first_order_momentum" parameter is deprecated')
 
+        self.gradient_accumulation_steps = gradient_accumulation_steps
+        self.device_iterations = device_iterations
+        self.inference_device_iterations = inference_device_iterations
+        self.optimizer_state_offchip = optimizer_state_offchip
+        self.replicated_tensor_sharding = replicated_tensor_sharding
+        self.auto_loss_scaling = auto_loss_scaling
         self.enable_half_partials = enable_half_partials
         self.executable_cache_dir = executable_cache_dir
         self.embedding_serialization_factor = embedding_serialization_factor
         self.recompute_checkpoint_every_layer = recompute_checkpoint_every_layer
         self.output_mode = output_mode
+
         # TODO: remove this if unnecessary.
         self.execute_encoder_on_cpu_for_generation = kwargs.pop("execute_encoder_on_cpu_for_generation", False)
 
@@ -469,7 +503,7 @@ class IPUConfig(BaseConfig):
                     f" same number of IPUs as {ipus_per_replica_mode_str}={self._ipus_per_replica}"
                 )
 
-            # layers_per_ipu must have the same length as ipus per replica.
+            # Layers_per_ipu must have the same length as ipus per replica.
             # If wildcards are present in layers_per_ipu, let the call to `model.parallelize`
             # handle the validation
             if -1 not in self._layers_per_ipu and len(self._layers_per_ipu) != self._ipus_per_replica:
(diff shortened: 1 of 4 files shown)


optimum/graphcore/ipu_configuration.py
```diff
@@ -178,6 +209,74 @@ def __init__(self, **kwargs):
                 matmul_proportion = 0.2
             setattr(self, f"{mode}_matmul_proportion", kwargs.pop(f"{mode}_matmul_proportion", matmul_proportion))
 
+            # mode based layer configurations
+            # serialized_linear_splits_per_ipu and serialized_embedding_splits_per_ipu defaults
+            # depend on the model architecture
```
HMellor, line 214: These defaults don't seem to depend on the model architecture? Should this commend be moved to the validation loop below?

optimum/graphcore/ipu_configuration.py
```diff
@@ -178,6 +209,74 @@ def __init__(self, **kwargs):
...
+                mode_layer_config = f"{mode}_serialized_{layer}_splits_per_ipu"
+                mode_layer_splits_per_ipu = getattr(self, mode_layer_config)
+                mode_layer_serialization_factor = getattr(self, f"{mode}_{layer}_serialization_factor")
+                if mode_layer_splits_per_ipu is not None:
+                    # serialized_{layer}_splits_per_ipu must be of type List[int>=0]
+                    if (
+                        not isinstance(mode_layer_splits_per_ipu, list)
+                        or not all(isinstance(x, int) for x in mode_layer_splits_per_ipu)
+                        or any(x < 0 for x in mode_layer_splits_per_ipu)
+                        or sum(mode_layer_splits_per_ipu) < 1
```
HMellor, line 248: This might just be personal preference but I find this version of the condition easier to read:
```suggestion
                    if not (
                        isinstance(mode_layer_splits_per_ipu, list)
                        and all(isinstance(x, int) for x in mode_layer_splits_per_ipu)
                        and all(x >= 0 for x in mode_layer_splits_per_ipu)
                        and sum(mode_layer_splits_per_ipu) > 0
```

optimum/graphcore/ipu_configuration.py
```diff
@@ -190,36 +230,68 @@ def contents_geq_value_validator(
...
         if value not in allowed_values:
             raise ValueError(
                 f"`IPUConfig` attribute `output_mode` can only take values in"
                 f" {allowed_values}. You provided: {value=}"
             )
 
-    attribute_validators[output_mode_validator] = {"output_mode"}
+    attribute_validators[_output_mode_validator] = {"output_mode"}
+
+    def _serialized_layer_splits_per_ipu_validator(name: str, value: str):
```
katalinic-gc, line 267: incorrect type hint for `value`

optimum/graphcore/modeling_utils.py
```diff
@@ -509,6 +554,90 @@ def forward(self, x):
...
+        self,
+        linear: torch.nn.Linear,
+        serialization_factor: int,
+    ) -> None:
+        super().__init__()
+
+        self.in_features = linear.in_features
+        self.out_features = linear.out_features
+        if self.in_features % serialization_factor != 0:
+            raise ValueError(f"{linear.in_features=} must be divisible by {serialization_factor=}")
```
katalinic-gc, line 580: is this error msg correct?

optimum/graphcore/modeling_utils.py
```diff
@@ -509,6 +554,90 @@ def forward(self, x):
...
+
+        self.split_size = int(self.in_features / serialization_factor)
```
katalinic-gc, line 582: if both of these are int, can you just do `self.in_features // serialization_factor`?

optimum/graphcore/modeling_utils.py
```diff
@@ -509,6 +554,90 @@ def forward(self, x):
...
+        if linear.bias is not None:
+            self.bias = torch.nn.Parameter(torch.zeros_like(linear.bias))
+            with torch.no_grad():
+                self.bias.copy_(linear.bias.detach())
+
+    def forward(self, x):
+        # each linear layer processes a partition of the input
+        out = None
+        for i, split_linear_layer in enumerate(self.split_linear_layers):
+            h = split_linear_layer(x[:, :, i * self.split_size : (i + 1) * self.split_size])
```
katalinic-gc, line 601: I think this should be changed to `x[..., i * self.split_size ...]` instead of `x[:, :, i * self.split_size ... ]`, x could be 2D as well?

optimum/graphcore/modeling_utils.py
```diff
@@ -509,6 +554,90 @@ def forward(self, x):
         return output
 
 
+class SplitProjection(torch.nn.Module):
```
jimypbr, line 557: Can you add a test that this gets same result as Linear?

tests/test_ipu_configuration.py
```diff
@@ -206,41 +206,84 @@ def test_layers_per_ipu(self):
         self.assertEqual(layer_ipu, [0, 1, 1, 2, 2, 3, 3])
 
     def test_execution_mode_specific_options(self):
```
jimypbr, line 208: this test class has ballooned in size. This needs to be broken up into different classes and smaller methods.

optimum/graphcore/modeling_utils.py
kundaMwiza: Yeah, the serialisation/splitting is along the reduction dimension

tests/test_modeling_utils.py
```diff
@@ -0,0 +1,165 @@
+import unittest
+from abc import abstractmethod
+
+import pytest
+import torch
+
+import poptorch
```
kundaMwiza, line 7: @jimypbr added to test the serialized layers. Perhaps we have something like this elsewhere?

optimum/graphcore/modeling_utils.py
```diff
@@ -541,6 +586,93 @@ def forward(self, x):
...
+
+    def to_model(self):
+        """
+        Merge the sub linear layers into one
+
+        Returns:
+            `nn.Linear` layer
+        """
+        dtype = self.split_linear_layers[0].weight.dtype
+        layer = nn.Linear(self.in_features, self.out_features, bias=False if self.bias is None else True)
```
katalinic-gc, line 654: nit: `bias=self.bias is not None`

katalinic-gc (approved):

optimum/graphcore/ipu_configuration.py
```diff
@@ -161,14 +194,18 @@ def __get__(self, obj, objtype=None):
...
 
     # Create a mapping of attributes to their list of validation functions
     attribute_validators = defaultdict(list)
+    _layers_per_ipu = ManagedAttribute("layers_per_ipu")
+    _ipus_per_replica = ManagedAttribute("ipus_per_replica")
+    _matmul_proportion = ManagedAttribute("matmul_proportion")
+    _projection_serialization_factor = ManagedAttribute("projection_serialization_factor")
+    _serialized_projection_splits_per_ipu = ManagedAttribute("serialized_projection_splits_per_ipu")
+    _embedding_serialization_factor = ManagedAttribute("embedding_serialization_factor")
+    _serialized_embedding_splits_per_ipu = ManagedAttribute("serialized_embedding_splits_per_ipu")
```
HMellor, line 206: Why the change in location?

optimum/graphcore/ipu_configuration.py
```diff
@@ -111,10 +111,43 @@ class IPUConfig(BaseConfig):
             Same as `matmul_proportion` for inference only.
         enable_half_partials (`bool`, *optional*, defaults to `True`):
             If `True`, sets the data type of partial results for matrix multiplication and convolution operators to float16.
-        embedding_serialization_factor (`int`, *optional*, defaults to 1):
+        projection_serialization_factor (`int`, *optional*, defaults to 1 if `serialized_projection_splits_per_ipu` is `None`):
```
HMellor, line 114: Would it make sense to leave the embedding option at the top as it appears "first" in the models this will be used in?

optimum/graphcore/ipu_configuration.py
```diff
@@ -317,15 +401,22 @@ def check_and_set_replication_factor(attr_name, attr, default=False):
         if "enable_half_first_order_momentum" in kwargs:
             warnings.warn('The "enable_half_first_order_momentum" parameter is deprecated')
 
+        self.gradient_accumulation_steps = gradient_accumulation_steps
+        self.device_iterations = device_iterations
+        self.inference_device_iterations = inference_device_iterations
+        self.optimizer_state_offchip = optimizer_state_offchip
+        self.replicated_tensor_sharding = replicated_tensor_sharding
+        self.auto_loss_scaling = auto_loss_scaling
```
HMellor, line 409: Why the change in location?

optimum/graphcore/ipu_configuration.py
```diff
@@ -419,20 +509,51 @@ def validate_ipu_config(self):
                 matmul_proportion_mode_str = self._get_managed_attr_mode_name("matmul_proportion")
                 raise IncompatibleIPUConfigError(
                     f"{matmul_proportion_mode_str}={self._matmul_proportion} should use the"
-                    f" same number of IPUs as {ipus_per_replica_mode_str}={self._ipus_per_replica}"
+                    f" same number of IPUs as {ipus_per_replica_mode_str}={self._ipus_per_replica}."
                 )
 
-            # layers_per_ipu must have the same length as ipus per replica.
+            # Layers_per_ipu must have the same length as ipus per replica.
```
HMellor, line 515: This is a variable name, it shouldn't be capitalised
```suggestion
            # layers_per_ipu must have the same length as ipus per replica.
```

optimum/graphcore/ipu_configuration.py
```diff
@@ -419,20 +509,51 @@ def validate_ipu_config(self):
...
+                    )
+
+                # Serialized layer splits per ipu pipeline must have the same pipeline length
+                # as the number of ipus per replica
+                if mode_layer_splits_per_ipu and len(mode_layer_splits_per_ipu) != self._ipus_per_replica:
+                    raise ValueError(
+                        f"{mode_layer_splits_per_ipu_str}={mode_layer_splits_per_ipu}"
+                        f" should use the same number of IPUs as {ipus_per_replica_mode_str}={self._ipus_per_replica}."
+                    )
+
```
HMellor, line 554: Could/should these be added as `attribute_validators`?

optimum/graphcore/ipu_configuration.py
```diff
@@ -419,20 +509,51 @@ def validate_ipu_config(self):
...
         self.mode = old_mode
+        return self
```
HMellor, line 556: Why return self? The return value doesn't appear to be used anywhere

optimum/graphcore/ipu_configuration.py
kundaMwiza: The issue with introducing attribute validators that compare the values of more than one attribute is ensuring that the order of validation is correct. If `*_splits_per_ipu` depends on `*ipus_per_replica` then we must make sure that `*ipus_per_replica` is validated first and it does not also depend on `*_splits_per_ipu`. If there is a circular dependency the only way to resolve attribute value conflicts would be to construct another `IPUConfig`

optimum/graphcore/ipu_configuration.py
kundaMwiza: I added that to support chaining method calls if `validate_ipu_config` is called elsewhere e.g. trainer.train() { .. ipu_config._validate_ipu_config() }

optimum/graphcore/ipu_configuration.py
kundaMwiza: Do you mean why does the order not match the docstring/ `__init__` parameter order? I placed them together because they don't have any special initialisation

HMellor: I was just wondering why these options were moved to this location

tests/test_modeling_utils.py
```diff
@@ -0,0 +1,165 @@
...
+#             options.deviceIterations(len(serialized_embedding_splits_per_ipu))
+
+#             other_evaluator = PoptorchEvaluator(
+#                 serialized_embedding.parallelize(serialized_embedding_splits_per_ipu),
+#                 options,
+#             )
+#         else:
+#             other_evaluator = PytorchEvaluator(serialized_embedding)
+
+#         return evaluator, other_evaluator, dataset
```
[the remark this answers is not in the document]
kundaMwiza: Another PR is needed to refactor `SerializedEmbedding`to use the to/from_model design. I can remove the code

optimum/graphcore/ipu_configuration.py
```diff
@@ -111,10 +111,44 @@ class IPUConfig(BaseConfig):
...
+            serialize the projection layer into a set of individual linear layers that can be optionally placed on different IPUs.
+            Nothing happens if `projection_serialization_factor = 1`. If `projection_serialization_factor > 1`,
+            the `torch.nn.Linear` layer is replaced by a `optimum.graphcore.modeling_utils.SplitProjection` layer
+            if `serialized_projection_splits_per_ipu` is provided and the linear layer's weights are not tied to another layer.
+            Otherwise it is replaced by a `optimum.graphcore.modeling_utils.SerializedLinear` layer.
+            Note: only one of `projection_serialization_factor` or `serialized_projection_splits_per_ipu` should be set.
+        inference_projection_serialization_factor (`int`, *optional*, defaults to 1 if `inference_serialized_projection_splits_per_ipu` is `None`):
+            Same as `projection_serialization_factor` but for inference only.
+        serialized_projection_splits_per_ipu (`List[int]`, *optional*, defaults to None):
+            Analogoues to `serialized_embedding_splits_per_ipu`.
```
jimypbr, line 148: spelling

jimypbr (approved):
LGTM

PR #356, opened 2023-04-24
Fix `test-pipelines.yml` and add host memory profiling to `RUN_SLOW=true` workflows

HMellor:
# What does this PR do?

Since the tests are parallelised, you can have `n` models compiling at the same time. It's possible that this can cause the CI runner to use all available memory and start using swap, which slows the runner down dramatically.

This PR should prevent that from happening in `test-pipeliens.yml` and add some host memory monitoring so we can identify if it happens again in the future.

## Before submitting


HMellor:
See https://github.com/huggingface/optimum-graphcore/actions/runs/4786164311 for an example memory profile artefact

jimypbr (approved):

PR #357, opened 2023-04-24
On device text generation POC for greedy search

katalinic-gc:
# What does this PR do?

Enabled via `on_device_generation_steps=on_device_generation_steps` kwarg to `model.parallelize`. Assumes KV caching is enabled, so will only work with Whisper for now.

There is an outstanding task to reconcile `generation_step` and `absolute_step` aka `cur_len - 1` usage since self-attn KV caching currently does not support a context of more than a single token.

## Before submitting


optimum/graphcore/generation/utils.py
```diff
@@ -1482,3 +1415,162 @@ def beam_sample(
...
+                new_max_length = max_length - on_device_generation_steps
+                logger.info(
+                    f"Temporarily adapting `max_length` from {max_length} to {new_max_length} for on device generation."
+                )
+                stopping_criterion = copy.deepcopy(stopping_criterion)
+                stopping_criterion.max_length = new_max_length
+            adapted_stopping_criteria.append(stopping_criterion)
+        return StoppingCriteriaList(adapted_stopping_criteria)
+
+    def _prepare_inputs_for_on_device_generation(self, model_inputs, on_device_generation_steps, batch_size):
```
jimypbr, line 1454: add a comment to say what this function is doing?

jimypbr (commented):
Looks great. Just add some comments and docstrings in a couple places that seemed complicated

jimypbr (approved):

PR #361, opened 2023-04-25
Update `actions/*` to fix Node.js warnings

HMellor:
# What does this PR do?

Should remove warnings in jobs caused by outdated version of Node.js used in `actions/checkout@v2` and `actions/setup-python@v2`

## Before submitting


jimypbr:
Now showing warning:
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

PR #364, opened 2023-04-28
Add typeguard to check attribute types on construction

kundaMwiza:
# What does this PR do?

- Adds `typeguard.typecheck` decorator to ensure method input and return types are adhered to. In particular, this ensures that the user provides the correct type to each attribute on construction.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -17,9 +17,10 @@ import copy
 import json
 import warnings
 from tempfile import NamedTemporaryFile
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, TypeVar, TypeAlias, List, Tuple
 
 import torch
+import typeguard
 
 import popart
 import poptorch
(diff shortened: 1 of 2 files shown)


optimum/graphcore/ipu_configuration.py
```diff
@@ -34,6 +35,13 @@
 ALLOWED_POD_TYPES = ["pod4", "pod8", "pod16", "pod32", "pod64"]
 
 
+class IPUConfigAttributeTypes:
+    T = TypeVar("T")
+    Type: TypeAlias = Union[T, Dict[str, T]]
+    OptionalType = Union[Optional[T], Dict[str, Optional[T]]]
```
HMellor, line 42: Maybe this? Or do we need to specify that some `T`s in a dict can be None too?
```suggestion
    OptionalType = Union[T, Dict[str, T], None]
```

optimum/graphcore/ipu_configuration.py
```diff
@@ -146,46 +150,118 @@ def __get__(self, obj, objtype=None):
...
+    replication_factor = ManagedAttribute()
+    device_iterations = ManagedAttribute()
+    embedding_serialization_factor = ManagedAttribute()
+    linear_serialization_factor = ManagedAttribute()
+    serialized_embedding_splits_per_ipu = ManagedAttribute()
+    serialized_linear_splits_per_ipu = ManagedAttribute()
+
+    def __init__(
+        self,
+        seed: IPUConfigAttributeTypes.OptionalType[int] = None,
```
HMellor, line 168: These are very long type hints and will take up a lot of space in the docs (if we want to use this typing class we'd also have to add it to the documentation. 

Do we need the optional versions of types? If they have default values (which they all do), then they're all already implicitly optional, right?

optimum/graphcore/ipu_configuration.py
kundaMwiza: Its possible that they are None. e.g. `inference_ipus_per_replica = None` by default

PR #366, opened 2023-05-02
Refactor: prefix IPUConfig `ManagedAttribute`s instead of overloading user provided attributes

kundaMwiza:
# What does this PR do?

The use of IPUConfig `ManagedAttribute`s currently involves the use of three attributes: `attribute`, `training_attribute` and `inference_attribute` (e.g. `layers_per_ipu`, `training_layers_per_ipu` and `inference_layers_per_ipu`). `attribute` is overrided internally to be a descriptor that returns `training_attribute` or `inference_attribute` depending on the `IPUConfig` mode. Since `attribute` is a descriptor, outside of the `Trainer` and `Pipeline` api, it is not guaranteed that the behaviour of the descriptor `attribute` will be consistent with the user expectation since the `IPUConfig` mode has to be set correctly.  This PR moves  `attribute` to a  descriptor `_attribute` that is handled internally, removing the need for the user to be concerned with the `IPUConfig` mode behaviour.

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -104,7 +104,7 @@ class IPUConfig(BaseConfig):
               - matrix multiplication
               - embedding lookups
               - indexing operations
-        inference_matmul_proportion (`List[int]`):
+        inference_matmul_proportion (`List[float]` or `float`):
             Same as `matmul_proportion` for inference only.
         enable_half_partials (`bool`, *optional*, defaults to `True`):
             Whether the data type of partial results for matrix multiplication and convolution operators should be
@@ -418,7 +418,13 @@ class IPUConfig(BaseConfig):
             ipus_per_replica_mode_str = self._get_managed_attr_mode_name("ipus_per_replica")
 
             # len(matmul_proportion) must equal ipus_per_replica
-            if isinstance(self.matmul_proportion, list) and len(self.matmul_proportion) != self.ipus_per_replica:
+            # if matmul_proportion = [proportion], _to_options will replicate
+            # its value to [proportion] * ipus_per_replica
+            if (
+                isinstance(self.matmul_proportion, list)
+                and len(self.matmul_proportion) != self.ipus_per_replica
+                and len(self.matmul_proportion) > 1
+            ):
                 matmul_proportion_mode_str = self._get_managed_attr_mode_name("matmul_proportion")
                 raise IncompatibleIPUConfigError(
                     f"{matmul_proportion_mode_str}={self.matmul_proportion} should use the"
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -252,21 +252,25 @@ class IPUConfig(BaseConfig):
 
         self.layers_per_ipu = layers_per_ipu
         self.inference_layers_per_ipu = inference_layers_per_ipu if inference_layers_per_ipu else self.layers_per_ipu
-        
+
         self.ipus_per_replica = ipus_per_replica if ipus_per_replica else len(self.layers_per_ipu)
         # If ipus_per_replica is default, recalculate ipus_per_replica from inference_layers_per_ipu instead
         fallback_ipus_per_replica = self.ipus_per_replica
         if fallback_ipus_per_replica == len(self.layers_per_ipu):
             fallback_ipus_per_replica = len(self.inference_layers_per_ipu)
-        self.inference_ipus_per_replica = inference_ipus_per_replica if inference_ipus_per_replica else fallback_ipus_per_replica
-        
+        self.inference_ipus_per_replica = (
+            inference_ipus_per_replica if inference_ipus_per_replica else fallback_ipus_per_replica
+        )
+
         self.matmul_proportion = matmul_proportion
         # If matmul_proportion is a list and its length is not equal to inference_ipus_per_replica, use the
         # default float value for matmul_proportion instead
-        fallback_matmul_proportion = self.matmul_proportion 
+        fallback_matmul_proportion = self.matmul_proportion
         if isinstance(self.matmul_proportion, list) and len(self.matmul_proportion) != self.inference_ipus_per_replica:
             fallback_matmul_proportion = 0.2
-        self.inference_matmul_proportion = inference_matmul_proportion if inference_matmul_proportion else fallback_matmul_proportion
+        self.inference_matmul_proportion = (
+            inference_matmul_proportion if inference_matmul_proportion else fallback_matmul_proportion
+        )
 
         def check_and_set_replication_factor(attr_name, attr, default=False):
             if isinstance(attr, int):
@@ -328,7 +332,7 @@ class IPUConfig(BaseConfig):
     @property
     def mode(self) -> str:
         return self._mode
-    
+
     @mode.setter
     def mode(self, value):
         if value not in self.modes:
@@ -445,7 +449,7 @@ class IPUConfig(BaseConfig):
 
         opts = Options()
         opts.autoRoundNumIPUs(True)
-        opts.replicationFactor(self.replication_factor)
+        opts.replicationFactor(self.inference_replication_factor if for_inference else self.replication_factor)
         opts.deviceIterations(self.inference_device_iterations if for_inference else self.device_iterations)
 
         if not for_inference:
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -418,27 +418,21 @@ class IPUConfig(BaseConfig):
             ipus_per_replica_mode_str = self._get_managed_attr_mode_name("ipus_per_replica")
 
             # len(matmul_proportion) must equal ipus_per_replica
-            # if matmul_proportion = [proportion], _to_options will replicate
-            # its value to [proportion] * ipus_per_replica
-            if (
-                isinstance(self.matmul_proportion, list)
-                and len(self.matmul_proportion) != self.ipus_per_replica
-                and len(self.matmul_proportion) > 1
-            ):
+            if isinstance(self._matmul_proportion, list) and len(self._matmul_proportion) != self._ipus_per_replica:
                 matmul_proportion_mode_str = self._get_managed_attr_mode_name("matmul_proportion")
                 raise IncompatibleIPUConfigError(
-                    f"{matmul_proportion_mode_str}={self.matmul_proportion} should use the"
-                    f" same number of IPUs as {ipus_per_replica_mode_str}={self.ipus_per_replica}"
+                    f"{matmul_proportion_mode_str}={self._matmul_proportion} should use the"
+                    f" same number of IPUs as {ipus_per_replica_mode_str}={self._ipus_per_replica}"
                 )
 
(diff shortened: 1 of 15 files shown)


optimum/graphcore/ipu_configuration.py
```diff
@@ -133,50 +129,51 @@ def __init__(self, attr) -> None:
...
+    _device_iterations = ManagedAttribute("device_iterations")
 
     def __init__(self, **kwargs):
         self.seed = kwargs.pop("seed", None)
 
         # Default mode to `training`
-        self.mode = "training"
+        self.train()
 
         # Get execution mode agnostic arguments
```
HMellor, line 158: This statement is no longer true

optimum/graphcore/ipu_configuration.py
```diff
@@ -133,50 +129,51 @@ def __init__(self, attr) -> None:
...
+        self.layers_per_ipu = kwargs.pop("layers_per_ipu", [-1])
+        self.ipus_per_replica = kwargs.pop("ipus_per_replica", len(self.layers_per_ipu))
+
+        self.inference_layers_per_ipu = kwargs.pop("inference_layers_per_ipu", self.layers_per_ipu)
+
+        # If ipus_per_replica is default, recalculate ipus_per_replica from inference_layers_per_ipu instead
+        ipus_per_replica = self.ipus_per_replica
+        if ipus_per_replica == len(self.layers_per_ipu):
+            ipus_per_replica = len(self.inference_layers_per_ipu)
+        self.inference_ipus_per_replica = kwargs.pop("inference_ipus_per_replica", ipus_per_replica)
```
HMellor, line 168:
```suggestion
        self.inference_layers_per_ipu = kwargs.pop("inference_layers_per_ipu", self.layers_per_ipu)
        self.inference_ipus_per_replica = kwargs.pop("inference_ipus_per_replica",len(self.inference_layers_per_ipu))
```

optimum/graphcore/ipu_configuration.py
kundaMwiza: Why?

optimum/graphcore/ipu_configuration.py
```diff
@@ -133,50 +129,51 @@ def __init__(self, attr) -> None:
...
+
+        # If matmul_proportion is a list and its length is not equal to inference_ipus_per_replica, use the
```
HMellor, line 170: I would move this comment down 2 lines so that it immediately precedes what it is referring to

optimum/graphcore/ipu_configuration.py
```diff
@@ -133,50 +129,51 @@ def __init__(self, attr) -> None:
...
+    # `self.mode` ("training" by default). For example `_layers_per_ipu`
+    # switches between `layers_per_ipu` and `inference_layers_per_ipu`
     modes = ("training", "inference")
-    layers_per_ipu = ManagedAttribute("layers_per_ipu")
-    ipus_per_replica = ManagedAttribute("ipus_per_replica")
-    matmul_proportion = ManagedAttribute("matmul_proportion")
+    _layers_per_ipu = ManagedAttribute("layers_per_ipu")
+    _ipus_per_replica = ManagedAttribute("ipus_per_replica")
+    _matmul_proportion = ManagedAttribute("matmul_proportion")
+    _device_iterations = ManagedAttribute("device_iterations")
```
HMellor, line 150: Was this added by accident? I don't see any reference to it elsewhere in this PR

optimum/graphcore/ipu_configuration.py
```diff
@@ -200,23 +197,39 @@ def __init__(self, **kwargs):
...
+    def mode(self) -> str:
+        return self._mode
+
+    @mode.setter
+    def mode(self, value: str) -> None:
+        if value not in self.modes:
+            raise ValueError(
+                f"`IPUConfig` mode can only be one of {self.modes}. Use the `train` and `eval` methods to avoid error."
+            )
+        self._mode = value
```
HMellor, line 225: Do we want the user to be able to set the mode this way? Writing to this property can be controlled by us using the `train()` and `eval()` methods

kundaMwiza: internally we use 
old_mode = self.mode
..
self.mode = old_mode

just thought its easier than conditioning on the value of old_mode

optimum/graphcore/ipu_configuration.py
kundaMwiza: I added it since it is a mode based attribute even though it is not used..yet?

optimum/graphcore/ipu_configuration.py
```diff
@@ -200,23 +197,39 @@ def __init__(self, **kwargs):
...
 
         # TODO: remove this if unnecessary.
         self.execute_encoder_on_cpu_for_generation = kwargs.pop("execute_encoder_on_cpu_for_generation", False)
 
+    def _get_managed_attr_mode_name(self, attr: str) -> str:
+        if hasattr(self, attr) and hasattr(self, f"inference_{attr}") and hasattr(self, f"_{attr}"):
+            return attr if self.mode == "training" else f"inference_{attr}"
+        else:
+            warnings.warn(f"{attr} is not a `ManagedAttribute`. Returning modeless name: {attr}")
+            return attr
```
HMellor, line 213: This function could be moved inside `ManagedAttribute` as a property called `full_name` and then it could be used like this:
```
    f"Invalid values in {layers_per_ipu.full_name}. {layers_per_ipu.full_name}={layers_per_ipu}"
)
```
Instead of creating a temporary string.

kundaMwiza: The name of the attribute depends on the mode

optimum/graphcore/modeling_utils.py
HMellor, line 1: With the suggestion made in `ipu_configuration.py` the `*_mode_str` parts wouldn't be necessary

HMellor (changes requested):
Made a few comments about suggested changes

optimum/graphcore/ipu_configuration.py
HMellor: Because `__init__` no longer takes mode agnostic arguments

optimum/graphcore/ipu_configuration.py
HMellor: If it isn't going to be get or set it's addition should be delayed to a PR that implements its usage too

optimum/graphcore/ipu_configuration.py
HMellor: Where is it use like this?

optimum/graphcore/ipu_configuration.py
HMellor: Exactly, and the `ManagedAttribute` knows the mode, which is how it returns the correct value. I'm suggesting we use exactly the same mechanism to return the correct name

optimum/graphcore/ipu_configuration.py
kundaMwiza: For example ipu_configuration.py:_to_options

optimum/graphcore/ipu_configuration.py
kundaMwiza: By mode agnostic arguments do you mean what `layers_per_ipu` was previously, or arguments for parameters like `gradient_accumulation_count`?

optimum/graphcore/ipu_configuration.py
HMellor: Ah, of course

optimum/graphcore/ipu_configuration.py
HMellor: Good point, so we still need to pass the mode to this function. Can we still make it a method of the `ManagedAttribute` class for DRY reasons as it already contains the `if else` logic for adding `inference_`?

kundaMwiza: This would make the syntax self.ipu_config.ManagedAttribute._get_attr_name(self.ipu_config.mode, "layers_per_ipu") because `self`=`ManagedAttribute`instance cannot be passed easily to `_get_attr_name` and so the `static`/ `class` method wouldn't know which attribute is being referred to

HMellor: Wouldn't it be `self.ipu_config.layers_per_ipu.full_name(self.ipu_config.mode)`?

kundaMwiza: Each attribute access from left to right returns the attribute if it exists, so on calling `full_name(..)`, that is called on the List[int] from `_layers_per_ipu`

kundaMwiza: The problem is roughly that `getattr(ipu_config, "_layers_per_ipu")` is internally replaced by `getattr(ipu_config, "_layers_per_ipu").__get__()` since `_layers_per_ipu` is a descriptor

optimum/graphcore/ipu_configuration.py
HMellor: Yeah originally this comment was for what `layers_per_ipu` was previously (i.e. not `training_` or `inference_`).

optimum/graphcore/ipu_configuration.py
HMellor: Ah of course, my mistake. You can ignore this comment

optimum/graphcore/ipu_configuration.py
```diff
@@ -133,50 +129,45 @@ def __init__(self, attr) -> None:
...
-                ipus_per_replica = len(getattr(self, f"{mode}_layers_per_ipu"))
-            setattr(self, f"{mode}_ipus_per_replica", kwargs.pop(f"{mode}_ipus_per_replica", ipus_per_replica))
-
-            # If matmul_proportion is a list and its length is not equal to {mode}_ipus_per_replica, use the
-            # default float value for matmul_proportion instead
-            if (isinstance(matmul_proportion, list) and  # fmt: skip
-                len(matmul_proportion) != getattr(self, f"{mode}_ipus_per_replica")):  # fmt: skip
-                matmul_proportion = 0.2
-            setattr(self, f"{mode}_matmul_proportion", kwargs.pop(f"{mode}_matmul_proportion", matmul_proportion))
+        self.layers_per_ipu = kwargs.pop("layers_per_ipu", [-1])
```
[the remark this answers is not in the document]
HMellor: Good spot, let's revert my suggestion

jimypbr:
I can do assignment:
```
ipu_config._ipus_per_config = 3
```
is that desired?

kundaMwiza:
> I can do assignment:
> 
> ```
> ipu_config._ipus_per_config = 3
> ```
> 
> is that desired?

@jimypbr The IPUConfig is not a frozen class that prevents users from adding new attributes? Or do you mean assignment  to managed attributes like layers_per_ipu: ipu_config._layers_per_ipu = 5? _layers_per_ipu is a private attribute so the user shouldn't be interacting with it. If we want to make sure that internally we don't use _layers_per_ipu to set the value of a training/inference attribute, that may cause complications with attributes like replication_factor which has its value overrided depending on n_ipus. However, replication_factor currently is not a ManagedAttribute..

katalinic-gc (approved):

PR #368, opened 2023-05-05
Add IPU support for HF pipelines to Whisper 

paolot-gc:
# Add support for HF pipelines

This PR makes it possible to use the HF pipeline for Whisper on the IPU.

## Before submitting

### DIFF
--- a/optimum/graphcore/pipelines/__init__.py
+++ b/optimum/graphcore/pipelines/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import poptorch
 import torch
@@ -185,7 +185,13 @@ SUPPORTED_TASKS = {
         "type": "text",
     },
 }
-SUPPORTED_GENERATION_TASKS = {"summarization", "text-generation", "text2text-generation", "translation"}
+SUPPORTED_GENERATION_TASKS = {
+    "automatic-speech-recognition",
+    "summarization",
+    "text-generation",
+    "text2text-generation",
+    "translation",
+}
 SUPPORTED_SEQ2SEQ_GENERATION_TASKS = {"summarization", "text2text-generation", "translation"}
 
 NO_FEATURE_EXTRACTOR_TASKS = set()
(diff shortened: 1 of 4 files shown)


katalinic-gc:
For `automatic_speech_recognition.py` can you indicate which parts are actual modifications? E.g. today in `generation/utils.py` we have comments like `# Change: ...`, perhaps `# IPU Change: ...` is clearer. The file seems to be mostly a copy and paste of upstream so its unclear what can be re-used and where we need IPU specific patches.

optimum/graphcore/pipelines/__init__.py
```diff
@@ -351,6 +366,12 @@ def pipeline(
 
     if ipu_config is None and not isinstance(model, poptorch._poplar_executor.PoplarExecutor):
         ipu_config = SUPPORTED_TASKS[targeted_task]["default"]["ipu_config"]
+    
+    optimisations = dict(
```
jimypbr, line 370: nit: I think we should use american spelling in code. `optimizations`

optimum/graphcore/pipelines/automatic_speech_recognition.py
```diff
@@ -0,0 +1,492 @@
...
+                    "Seq2Seq speech recognition model requires either a "
+                    f"`input_features` or `input_values` key, but only has {model_inputs.keys()}"
+                )
+
+            # we need to pass `processed.get("attention_mask")` here since audio encoder
+            # attention mask  length is different from expected text decoder `encoder_attention_mask` length
+            # `generate` magic to create the mask automatically won't work, we basically need to help
+            # it here.
+            attention_mask = model_inputs.pop("attention_mask", None)
+            #IPU: For Whisper we call generate() without encoder_output nor attention_mask
```
katalinic-gc, line 378: I suspect there's a better way to do this but I need to have a look at the code before making suggestions.

optimum/graphcore/pipelines/__init__.py
```diff
@@ -360,13 +378,18 @@ def pipeline(
             "Using a pipeline without specifying a model name and revision in production is not recommended."
         )
         model = SUPPORTED_TASKS[targeted_task]["class"][0].from_pretrained(model_id, revision=revision)
```
jimypbr, line 380: this currently selects the 0th element from the "class" tuple. It should loop over the classes and `try` `from_pretrained` until it works. See the code from transformers that does the same: https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/base.py#L269

PR #371, opened 2023-05-06
Add attribute validation

kundaMwiza:
# What does this PR do?

- Add attribute value type checking 
- Add attribute value validation
- Add validation of ipu config values

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -14,11 +14,13 @@
 #  limitations under the License.
 
 import copy
+from functools import partial
 import json
 import warnings
 from tempfile import NamedTemporaryFile
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, List, get_type_hints, Type, Set, Sequence
 
+import typeguard
 import torch
 
 import popart
@@ -31,6 +33,10 @@ from poptorch import Options, OutputMode
 logger = logging.get_logger(__name__)
 
 IPU_CONFIG_NAME = "ipu_config.json"
+
+class InvalidIPUConfigError(ValueError):
+    "Exception raised when IPU config values are not coherent"
+    pass
 
 
 class IPUConfig(BaseConfig):
@@ -76,6 +82,11 @@ class IPUConfig(BaseConfig):
             Same as `layers_per_ipu` for training only.
         inference_layers_per_ipu (`List[int]`):
             Same as `layers_per_ipu` for inference only.
+        ipus_per_replica (`int`, *optional*, defaults to `len(layers_per_ipu)`):
+            Specifies the number of IPUs to use during training. This must be consistent with 
+            the number of IPUs used in `layers_per_ipu`.
+        inference_ipus_per_replica (`int`, *optional*, defaults to `len(inference_layers_per_ipu) if ipus_per_replica==len(layers_per_ipu) else ipus_per_replica):
+            Same as `ipus_per_replica` but for inference only.
 
         > Parameters for memory management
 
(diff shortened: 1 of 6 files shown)


optimum/graphcore/ipu_configuration.py
```diff
@@ -200,26 +305,125 @@ def __init__(self, **kwargs):
...
+        """
+        Tests coherence of `IPUConfig` attributes for all modes
+        in self.modes. For example if `matmul_proportion=[0.2, 0.2]`,
+        `ipus_per_replica` must have value 2.
+
+        Raises:
+            InvalidIPUConfigError: Raised if any `IPUConfig` attributes
+            are not coherent.
+        """
+        # import pdb; pdb.set_trace()
```
katalinic-gc, line 396: debugging leftover

optimum/graphcore/ipu_configuration.py
```diff
@@ -16,23 +16,33 @@
...
 
+# For container types check all items for type correctness
+# rather than just the first element
+typeguard._config.global_config.collection_check_strategy = typeguard.config.collection_check_strategy.ALL_ITEMS
 logger = logging.get_logger(__name__)
 
 IPU_CONFIG_NAME = "ipu_config.json"
 
 
+class InvalidIPUConfigError(ValueError):
```
katalinic-gc, line 41: Nit: there exists `IncompatibleIPUConfigError` in `modeling_utils.py`. Could you replace it with this `InvalidIPUConfigError` since they are basically the same thing?

katalinic-gc:
lgtm

optimum/graphcore/ipu_configuration.py
```diff
@@ -151,46 +164,139 @@ def __get__(self, obj, objtype=None):
...
+            raise ValueError(
+                f"`IPUConfig` attribute `output_mode` can only take values in"
+                f" {allowed_values}. You provided: {value=}"
+            )
+
+    attribute_validators[output_mode_validator] = {"output_mode"}
+
+    def __init__(
+        self,
+        seed: Optional[int] = None,
```
jimypbr, line 227: we should maybe reorder the parameters so they appear in the same order as the docstring.
just thinking about using this in a jupyter notebook. I'd rather more useful things appeared near the top. seed, auto_loss_scaling, executable_cache_dir shouldn't be the first things imo.

kundaMwiza: Currently this list is in the same order as the doctoring. Do you mean we should change the docstring so that seed, auto_loss_scaling, executable_cache_dir are placed at the end?

jimypbr: ah so it is github was hiding it from me. nvm then.

optimum/graphcore/ipu_configuration.py
```diff
@@ -200,26 +306,124 @@ def __init__(self, **kwargs):
...
+        """Override __setattr__ to include value type checking
+        and validation
+        """
+        attr_type = self._get_attribute_type(name)
+        try:
+            typeguard.check_type(value, attr_type)
+        except typeguard.TypeCheckError as e:
+            raise TypeError(
+                f"Setting `IPUConfig` attribute: {name}, type: {attr_type}."
+                f" with {value=}, type: {type(value)} is invalid."
```
jimypbr, line 377: This produces error message:
```
TypeError: Setting `IPUConfig` attribute: training_layers_per_ipu, type: typing.Union[typing.List[int], NoneType]. with value=2, type: <class 'int'> is invalid.
```

Capitalise `with` or remove full stop.

kundaMwiza: @jimypbr I've removed the full stop. This is the intended behaviour, no?

jimypbr:
This seems to work without error:
```
ipu_config = IPUConfig(inference_replication_factor=0.2)
```
The `inference_replication_factor` gets set to `1` in this case.

optimum/graphcore/ipu_configuration.py
jimypbr: yeah it was just grammar nit

jimypbr (approved):
LGTM

PR #379, opened 2023-05-12
Serialized linear improvements

jimypbr:
# What does this PR do?

Change `SerializedLinear` to use the `from_model`/`to_model` pattern and update all uses of it in the models to use these methods.

Fixes # (issue)

## Before submitting


katalinic-gc (approved):
lgtm

PR #380, opened 2023-05-15
Update ipu configuration tests

kundaMwiza:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/ipu_configuration.py
+++ b/optimum/graphcore/ipu_configuration.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 
 import copy
+from collections import defaultdict
 import json
 import warnings
 from functools import partial
@@ -163,9 +164,8 @@ class IPUConfig(BaseConfig):
     _ipus_per_replica = ManagedAttribute("ipus_per_replica")
     _matmul_proportion = ManagedAttribute("matmul_proportion")
 
-    # Create a mapping of attribute value validating functions to a set of attributes
-    # to be validated by that function
-    attribute_validators = dict()
+    # Create a mapping of attributes to their list of validation functions
+    attribute_validators = defaultdict(list)
 
     def contents_geq_value_validator(
         name: str, value: Union[float, int, Sequence], floor_value: Union[float, int]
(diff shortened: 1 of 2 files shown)


tests/test_ipu_configuration.py
```diff
@@ -37,7 +38,45 @@ def create_ipu_config() -> IPUConfig:
...
 
 
+def create_mode_ipu_config(test_attributes: Dict[str, Any], mode):
+    helper_config = IPUConfig()
+    helper_config.mode = mode
+    ipu_config = IPUConfig(
+        **{helper_config._get_managed_attr_mode_name(attr): value for attr, value in test_attributes.items()}
+    )
+    ipu_config.mode = mode
+    return ipu_config
```
HMellor, line 48: If `_get_managed_attr_mode_name` was a class method, could we remove `helper_config`?

tests/test_ipu_configuration.py
```diff
@@ -37,7 +38,45 @@ def create_ipu_config() -> IPUConfig:
...
+        # ipus_per_replica should equal to len(layers_per_ipu) if not provided
+        ipu_config = IPUConfig(layers_per_ipu=[1, 2, 3])
+        self.assertEqual(ipu_config.ipus_per_replica, 3)
+
+        # inference_matmul_proportion not specified but matmul_proportion is
+        ipu_config = IPUConfig(
+            layers_per_ipu=[1, 2, 3, 4],
+            matmul_proportion=[0.1, 0.2, 0.3, 0.4],
+        )
+        self.assertEqual(ipu_config.inference_matmul_proportion, [0.1, 0.2, 0.3, 0.4])
```
HMellor, line 62: This misses what was originally being tested. It is supposed to test that `matmul_proportion` _isn't_ copied if it's not compatible with `inference_layers_per_ipu`
```suggestion
        # inference_matmul_proportion not specified but matmul_proportion is
        ipu_config = IPUConfig(
            layers_per_ipu=[1, 2, 3, 4],
            inference_layers_per_ipu=[3, 7],
            matmul_proportion=[0.1, 0.2, 0.3, 0.4],
        )
        self.assertEqual(ipu_config.inference_matmul_proportion, 0.2)
```

tests/test_ipu_configuration.py
kundaMwiza: the method requires the `self.mode` unfortunately

tests/test_ipu_configuration.py
```diff
@@ -334,101 +324,126 @@ def test_split_encoder_decoder_ipu_config(self):
...
+        (
+            (attr, test_value, mode)
+            for mode in IPUConfig.modes
+            for attr, test_value in {
+                "layers_per_ipu": [1, 2, 3, 4],
+                "matmul_proportion": 0.6,
+                "ipus_per_replica": 4,
+            }.items()
+        )
+    )
```
HMellor, line 354: This uses the same values for both modes, so how can we differentiate which mode's values were returned?

kundaMwiza: On construction only one of attr / inference_attr is specialised. So getattr(ipu_config, "_{attr}") should equal the specialised value

kundaMwiza: Its essentially like this:
```
mode = "inference"
ipu_config = IPUConfig(inference_attr = value) if mode == "inference" else IPUConfig(attr = value)
ipu_config.mode = mode
self.assertEqual(getattr(ipu_config., f"_{attr}"), value)
```

optimum/graphcore/ipu_configuration.py
```diff
@@ -163,9 +164,8 @@ def __get__(self, obj, objtype=None):
     _ipus_per_replica = ManagedAttribute("ipus_per_replica")
     _matmul_proportion = ManagedAttribute("matmul_proportion")
 
-    # Create a mapping of attribute value validating functions to a set of attributes
-    # to be validated by that function
-    attribute_validators = dict()
+    # Create a mapping of attributes to their list of validation functions
+    attribute_validators = defaultdict(list)
```
HMellor, line 168: Why the list? As far as I can tell, each `attr` only has 1 validation function. Is this so it can be extended in future?

kundaMwiza: Yeah^

tests/test_ipu_configuration.py
HMellor: Would this work?
```    
def _get_managed_attr_mode_name(self, attr: str, mode: str = None) -> str:
        """
        Returns the attribute name that a ManagedAttribute descriptor is
        currently referring to
        """
        # Shallow check to ensure that the input attribute is actually
        # a managed attribute
        if hasattr(self, attr) and hasattr(self, f"inference_{attr}"):
            return attr if (mode or self.mode) == "training" else f"inference_{attr}"
        # return attr if its not a managed attribute
        return attr
```

optimum/graphcore/ipu_configuration.py
kundaMwiza: On main, its currently function -> Attribute set in the `attribute_validators` map, rather than attribute -> validators. Now __set__ just needs to check if the attribute is in `attribute_validators` rather than iterating over all validator functions and checking if an attribute is validated by that function

tests/test_ipu_configuration.py
kundaMwiza: `__init__` is required to be run in order to know if the attribute has both attr and inference_attr

HMellor: Ok, let's leave this as is. (feel free to mark this thread as resolved)

HMellor (approved):

PR #384, opened 2023-05-18
Only `parallelize(for_generation=True)` if pipeline task is in `SUPPORTED_GENERATION_TASKS`

HMellor:
# What does this PR do?

Fixes failing pipeline tests coming from index error in `fill-mask` pipeline

To validate:
```
RUN_SLOW=true pytest -k BartForConditionalGeneration tests/pipelines/test_pipelines_fill_mask.py -v
```

## Before submitting

### DIFF
--- a/tests/pipelines/test_pipelines_fill_mask.py
+++ b/tests/pipelines/test_pipelines_fill_mask.py
@@ -167,18 +167,14 @@ class FillMaskPipelineTests(unittest.TestCase, metaclass=PipelineTestCaseMeta):
             ipu_config=ipu_config,
             tokenizer=tokenizer,
         )
-        examples = [
-            f"This is another {tokenizer.mask_token} test",
-        ]
+        examples = [f"This is a {tokenizer.mask_token} test"]
         return fill_masker, examples
 
     def run_pipeline_test(self, fill_masker, examples):
         tokenizer = fill_masker.tokenizer
         model = fill_masker.model
 
-        outputs = fill_masker(
-            f"This is a {tokenizer.mask_token}",
-        )
+        outputs = fill_masker(examples[0])
         self.assertEqual(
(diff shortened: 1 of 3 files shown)


tests/pipelines/test_pipelines_fill_mask.py
```diff
@@ -202,7 +198,7 @@ def run_pipeline_test(self, fill_masker, examples):
             ],
         )
 
-        outputs = fill_masker([f"This is a {tokenizer.mask_token}", f"Another {tokenizer.mask_token} great test."])
+        outputs = fill_masker(examples * 2)
```
katalinic-gc, line 201: How come you are changing the test example here?

HMellor: Because currently this test is not using the `examples` provided by the function arguments. So I just changed the test cases so that they did. It's still testing `str, list(str), list(str, str)` like it was before.

katalinic-gc: Do you by any chance know when the test started failing? We didn't do a transformers update so the need to add postprocess seems like it would have been failing for quite a while 🤔

HMellor: You're right that this is probably the wrong location for the fix.

The issue is that the logits output from BART are of shape `[batch_size, 1, vocab_size]` (presumably the 1 is the number of mask tokens) but for other models like BERT the logits are of shape `[batch_size, sequence_length, vicab_size]`.

The added `postprocess` tweak works because for all other models the `expand` is a no-op, but for BART it corrects the output shape (by faking all the other logits). 

To truly solve this we need to make sure that BART outputs the logits for the full sequence.

katalinic-gc (approved):

PR #407, opened 2023-06-06
Enable generation tests

kundaMwiza:
# What does this PR do?

- update generation tests structure to mirror transformers 4.29.2
- update generation strategies in optimum/graphcore/generation/utils.py to support multiple eos token to mirror transformers 4.29.2
- add destroy method to mirror poptorch_model.destroy()

Fixes # (issue)

## Before submitting

### DIFF
--- a/tests/generation/test_utils.py
+++ b/tests/generation/test_utils.py
@@ -18,14 +18,12 @@ import inspect
 import unittest
 from typing import TYPE_CHECKING, List, Optional, Union
 
-import pytest
+import numpy as np
 
-from optimum.graphcore import IPUConfig
+from optimum.graphcore import IPUConfig, pipeline
 from optimum.graphcore.modeling_utils import to_pipelined
 from transformers import is_torch_available
 from transformers.testing_utils import require_torch, slow, torch_device
-
-from optimum.graphcore import IPUConfig, pipeline
 
 from ..test_modeling_common import floats_tensor, ids_tensor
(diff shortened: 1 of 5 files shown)


tests/generation/test_utils.py
```diff
@@ -1524,159 +1532,187 @@ def _check_sequence_inside_sequence(self, tensor_1, tensor_2):
...
+    def parallelize_model(cls, model: torch.nn.Module, ipu_config: IPUConfig = None, **generation_kwargs):
+        if ipu_config is None:
+            ipu_config = cls.default_ipu_config
+
+        model = to_pipelined(model, ipu_config)
+        GenerationIntegrationTests.test_model = model
+
+        def reparallelize(self, **generation_kwargs):
+            self.deparallelize()
+            self.parallelize(for_generation=True, use_cache=self.config.use_cache, **generation_kwargs)
```
katalinic-gc, line 1682: double checking, should this be config, or generation_config (which might be None) ? and similarly for other references / use_cache checks

katalinic-gc: Whilst above is true, there is a step afterwards where if a model has a generation config on the hub, the generation config will be updated https://github.com/gante/transformers/blob/f85e3b2f73ebeb36c02e83684f0465829db10edd/src/transformers/modeling_utils.py#L2494. If they can all be used for generation, then generation_config should be defined?

PR #425, opened 2023-06-19
Add eval parallelize kwargs to seq2seq trainer

katalinic-gc:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/trainer_seq2seq.py
+++ b/optimum/graphcore/trainer_seq2seq.py
@@ -59,6 +59,7 @@ class IPUSeq2SeqTrainer(IPUTrainer):
         optimizers: Tuple[torch.optim.Optimizer, torch.optim.lr_scheduler.LambdaLR] = (None, None),
         preprocess_logits_for_metrics: Optional[Callable[[torch.Tensor, torch.Tensor], torch.Tensor]] = None,
         force_to_pipelined: bool = False,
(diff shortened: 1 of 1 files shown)


optimum/graphcore/trainer_seq2seq.py
```diff
@@ -130,7 +137,7 @@ def _wrap_and_compile_model_for_evaluation(self, dataloader, prediction_loss_onl
 
         # reparallelize for generation
         self.model.deparallelize().ipu_config.eval()
-        self.model.parallelize(for_generation=True)
+        self.model.parallelize(for_generation=True, **self.eval_parallelize_kwargs)
```
kundaMwiza, line 140: The user can provide a generation config via IPUSeq2SeqTrainerArguments. If they do not, the default generation config for the model will be used. Can we use the generation config to populate the parallelize kwargs instead?

katalinic-gc: Whilst the generation config may in theory be used to carry the dict (it would not be an explicit field, and **kwargs get forwarded to generate(), which is too late), I don't think we should overload it for that. The generation config is used to control the behaviour of `generate()` (basically it just collected existing kwargs of generate into one config). This comes after model creation which in our case includes parallelization. Both model creation and `pipelines` currently have means of supplying some variant `parallelize_kwargs`, so I think it makes sense to continue being explicit.

PR #427, opened 2023-06-20
Add parallelize kwargs as an IPU config entry

katalinic-gc:
# What does this PR do?

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/pipelines/__init__.py
+++ b/optimum/graphcore/pipelines/__init__.py
@@ -216,7 +216,6 @@ def get_poplar_executor(
     ipu_config: Union[IPUConfig, str, dict] = None,
     fp16: bool = True,
     for_generation: bool = False,
-    **parallelize_kwargs,
 ) -> PreTrainedModel:
     ipu_config_arg = ipu_config
 
@@ -231,6 +230,8 @@ def get_poplar_executor(
     # that are different in training and inference
     ipu_config.eval()
 
+    parallelize_kwargs = ipu_config.inference_parallelize_kwargs
+
     ipu_config.inference_device_iterations = 1
     if not parallelize_kwargs.get("use_cond_encoder", False):
         ipu_config.inference_replication_factor = 1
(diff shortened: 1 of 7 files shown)


optimum/graphcore/trainer.py
```diff
@@ -1318,7 +1318,7 @@ def _load_best_model(self):
     def _load_state_dict_in_model(self, state_dict):
         self.model.deparallelize()
         load_result = self.model.load_state_dict(state_dict, strict=False)
-        self.model.parallelize()
+        self.model.parallelize(**self.model.ipu_config.parallelize_kwargs)
```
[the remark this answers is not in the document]
katalinic-gc: I only see this called via `_load_best_model` -> `_load_state_dict_in_model` which will be a training model. So checking the mode would mainly be to prevent future errors, and given it's only a single site, wouldn't it be simpler to just check the mode the model is running in?

optimum/graphcore/pipelines/__init__.py
```diff
@@ -243,7 +244,7 @@ def get_poplar_executor(
                 parallelize_kwargs["use_cache"] = True
             model.parallelize(for_generation=for_generation, **parallelize_kwargs)
```
kundaMwiza, line 245: This comment is not related to the aim of this PR but, shouldn't this be like so? Why do we gate on the model being an encoder-decoder type?
```
if isinstance(model, IPUGenerationMixin)
    if "use_cache" not in parallelize_kwargs and model.__class__ in MODELS_SUPPORTING_KV_CACHE:
          ... 
    else:
         model.parallelize(**parallelize_kwargs)
else:
    raise Error
```

kundaMwiza:
Pipelined models require an ipu config in order to parallelize. If kwargs is not populated on the parallelize call, can we defer to the parallelize kwargs in the ipu config?

```python
def parallelize(for_generation=False, **kwargs):
    kwargs = PipelineMixin.parallelize(kwargs)  
 ```

The above avoids having to reference the already provided ipu_config on parallelize:

```python
model = to_pipelined(model, ipu_config)
model.parallelize(**ipu_config._parallelize_kwargs)
```

optimum/graphcore/pipelines/__init__.py
katalinic-gc: Only models where we support the parallelize optimisations (which were basically for generation purposes) were for encoder-decoder architectures, so it was something put in at the beginning. This is especially the case with KV caching.

PR #429, opened 2023-06-22
Add group quantization for whisper

jimypbr:
# What does this PR do?

Adds group quantization custom op to optimum. Adds the functionality to whisper's linear layers.

Still to do:
- [x] Automate the custom op build into setup.py
- [x] Add quantisation options into IPUConfig or whisper parallelize kwargs.

Fixes # (issue)

## Before submitting


katalinic-gc (approved):

optimum/graphcore/quantization/group_quantize.py
```diff
@@ -0,0 +1,213 @@
...
+    @classmethod
+    def from_model(cls, linear: nn.Linear, num_groups: int):
+        w = linear.weight.data
+        bias = linear.bias
+        w_packed, w_scale, w_bias = group_quantize_compress(w, num_groups)
+        return cls(linear.in_features, linear.out_features, w_packed, w_scale, w_bias, bias)
+
+    def forward(self, input):
+        weight = group_quantize_decompress(self.w_packed.data, self.w_scale.data, self.w_bias.data, dtype=input.dtype)
+        output = input @ weight.t()
```
katalinic-gc, line 210: Could you change this to segment to be consistent with upstream linear, which calls `F.linear(input, self.weight, self.bias)`?

jimypbr: yep. I had that originally, but changed to this to debug. good catch.

PR #431, opened 2023-06-26
Support finetuning with LoRA

katalinic-gc:
# What does this PR do?

To enable it, apply below, which is basically identical to upstream.

```python
config = LoraConfig(
    r=16, 
    lora_alpha=32, 
    target_modules=["q_proj", "v_proj"], 
    lora_dropout=0.05, 
    bias="none"
)
model = get_peft_model(model, config)
model.print_trainable_parameters()
```

Some online finetuning walkthroughs also include
```python
    remove_unused_columns=False,  # required as the PeftModel forward doesn't have the signature of the wrapped model's forward
    label_names=["labels"],  # same reason as above
```

to the training args, e.g. `(IPU)Seq2SeqTrainingArguments`.

## Before submitting


notebooks/whisper_lora.ipynb
katalinic-gc, line 1: Notebook is WIP.

- Yet to decide on what dataset is most appropriate (some ASR datasets are quite large)
- I predict a request to actually show a generation or pipelines example at the end of the notebook, so I'll probably add something in

rrva:
@katalinic-gc I would like to use this to finetune whisper-large-v2 on a fairly large dataset (700k examples). Is this usable as is (if I fix the merge conflicts)? What steps from https://www.graphcore.ai/posts/fine-tune-openais-whisper-automatic-speech-recognition-asr-model would need to be different with this approach?

katalinic-gc:
> @katalinic-gc I would like to use this to finetune whisper-large-v2 on a fairly large dataset (700k examples). Is this usable as is (if I fix the merge conflicts)? What steps from https://www.graphcore.ai/posts/fine-tune-openais-whisper-automatic-speech-recognition-asr-model would need to be different with this approach?

It won't be usable for large on IPUs due to OOM. We are internally working on supporting that; if and when available, we'll announce it.

PR #436, opened 2023-06-27
Base models and notebooks for general IPU embeddings model

arsalanu:
Support for multiple embeddings models (e5, mpnet, t5 (pending))
Idea is to create a general embeddings class which supports the pool and norm stage of embeddings model. These models use encoder stages from various base models (bert, t5, mpnet) which we don't support the direct use of with pipelining. I've added these in this folder for now, but they may need to be in the source lib if integrated.

Seemed like not a great idea to create new pipelined classes for all of these models, as they inherit from existing models we support and would just be an extra couple of lines which are common amongst them, so makes more sense to have a generic wrapper for embeddings, which instantiates the encoder stages through AutoModel as currently shown in HF.

The inference setup is manually outlined, but may be more user-friendly to create a pipeline for these eventually.

MPNet modeling file only includes the base model and maskedLM model, not tested for pretraining, only for embeddings.

notebook will likely be restructured


arsalanu:
relies on #437 being merged first

notebooks/text_embeddings_models/config.py
```diff
@@ -0,0 +1,103 @@
...
+architectures = ['BertModel', 'MPNetModel', 'MPNetForMaskedLM', 'T5EncoderModel']
+
+def get_ipu_config(model_config, n_ipu, model_ipu, device_iterations, replication_factor=None, random_seed=None):
+    base_architecture = model_config.architectures[0]
+
+    if base_architecture not in architectures:
+        logger.error(f"Model config passed does not contain a supported architecture: {architectures}")
+        raise ValueError("Unsupported model architecture.")
+
+    if model_ipu not in [1,4]:
```
katalinic-gc, line 26: Can you rename this to `ipus_per_replica`? That's more aligned with the naming adopted in the `IPUConfig`.

notebooks/text_embeddings_models/text-embeddings-on-ipu.ipynb
katalinic-gc, line 1: Please clear notebook outputs

notebooks/text_embeddings_models/text-embeddings-on-ipu.ipynb
```diff
@@ -0,0 +1,783 @@
...
+    "* [Sentence-T5](https://arxiv.org/abs/2108.08877), which runs on a pre-trained T5 model encoder."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "6d591cc4-38ca-47ae-a52f-165b04deb0a8",
+   "metadata": {},
+   "source": [
+    "First, install the requirements for running this notebook:"
```
katalinic-gc, line 30: I guess the assumption here is that one already has our SDK and `optimum-graphcore` installed? I think this needs to be made clearer - ideally the assumption wouldn't be made at least for `optimum-graphcore` - as this notebook quite notably deviates from others (that are more aligned to the PS template).

katalinic-gc: An alternative would be that we land the source model changes (everything outside the `notebooks`) as a separate PR, then in this PR you instead install `optimum-graphcore` from source, pointing at the commit associated with the separate PR.

I would expect the combined call to install `optimum-graphcore+... sentence-transformers` to work fine without the torch cpu stuff. Also ideally `sentence-transformers` would be pinned as well, e.g. to the version you are using now.

arsalanu: will uncomment this in a PR after landing  + `sentence-transformers` is pinned and torch-cpu is removed

PR #437, opened 2023-06-27
Support T5EncoderModel for t5-based embedding models

alex-coniasse:
# What does this PR do?
Add support for the encoder-only model T5EncoderModel
used in t5 embeddings models such as sentence-transformers.
(related to #436)

Fixes # (issue)

## Before submitting


tests/ipu_config_t5Encoder.json
katalinic-gc, line 1: Please create a model card entry on the hub with the IPU config instead of adding the json file here as that is the approach taken for all other tests.

tests/utils.py
```diff
@@ -40,5 +41,17 @@
 }
 
 
+# Registered models that don't have a mapping in upstream transformers
+class T5EncoderConfig(T5Config):
+    pass
+
+
+MODEL_MAPPING_EXTRA = {T5EncoderConfig: T5EncoderModel}
+EXTRA_CONFIG_MAPPING = {"t5Encoder": T5EncoderConfig}
```
katalinic-gc, line 50: Nit, but could you change these new maps to use a consistent naming convention, preferably have `EXTRA` as a suffix.

tests/utils.py
```diff
@@ -40,5 +41,17 @@
 }
 
 
+# Registered models that don't have a mapping in upstream transformers
+class T5EncoderConfig(T5Config):
```
katalinic-gc, line 45: Could you move this to the `models/t5` folder instead? Upstream does similar where model specific stuff is grouped into appropriate subfolders, with configs placed into `models.t5.configuration_t5.py`.

alex-coniasse:
Thanks for the review @katalinic-gc , 
I'll work on publishing a model card then,

katalinic-gc (approved):

PR #438, opened 2023-06-29
SplitProjection layer: Add output channels serialization mode

kundaMwiza:
# What does this PR do?

- Support output channels matmul serialization mode.
- Sets this as the new default to match SerializedLinear

Fixes # (issue)

## Before submitting

### DIFF
--- a/optimum/graphcore/modeling_utils.py
+++ b/optimum/graphcore/modeling_utils.py
@@ -16,14 +16,13 @@ from __future__ import annotations
 import copy
 from typing import List, Optional, Tuple, Union
 
-from functools import partial
 import poptorch
 import torch
 import torch.nn.functional as F
 from torch import nn
-from transformers import PreTrainedModel
 
 from optimum.utils import logging
+from transformers import PreTrainedModel
 
 from .ipu_configuration import IncompatibleIPUConfigError, IPUConfig
 
@@ -622,7 +621,7 @@ class SplitProjection(torch.nn.Module):
         self,
         linear: torch.nn.Linear,
         serialization_factor: int,
-        serialization_mode = poptorch.MatMulSerializationMode.OutputChannels
+        serialization_mode=poptorch.MatMulSerializationMode.OutputChannels,
     ) -> None:
         super().__init__()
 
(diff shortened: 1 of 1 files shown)


optimum/graphcore/modeling_utils.py
```diff
@@ -611,51 +611,91 @@ class SplitProjection(torch.nn.Module):
...
+        else:
+            raise ValueError(
+                f"`SplitProjection` `serialization_mode` can only be {poptorch.MatMulSerializationMode.OutputChannels} or {poptorch.MatMulSerializationMode.ReducingDim}."
+                f" You provided: {serialization_mode=}"
+            )
 
         self.bias = None
         if linear.bias is not None:
             self.bias = torch.nn.Parameter(torch.zeros_like(linear.bias))
+            if linear.weight.dtype is torch.float16:
```
katalinic-gc, line 668: Is this necessary? `zeros_like` will create a tensor of same dtype as `linear.bias`. When would the dtypes of weight and bias of a linear layer be mismatched?

optimum/graphcore/modeling_utils.py
```diff
@@ -611,51 +611,91 @@ class SplitProjection(torch.nn.Module):
...
+            For matrix A (m by n) multiplied by matrix B (n by p):
+            * ReducingDim: Split across the reducing dimension (n).
+            * OutputChannels: Split across the output channels (dimension p).
     """
 
     def __init__(
         self,
         linear: torch.nn.Linear,
         serialization_factor: int,
+        serialization_mode=poptorch.MatMulSerializationMode.OutputChannels,
```
katalinic-gc, line 624: I know this layer is only used in most recent models, did you check it won't break current usage as this changes the default from ReducingDim to OutputChannels?

optimum/graphcore/modeling_utils.py
```diff
@@ -611,51 +611,91 @@ class SplitProjection(torch.nn.Module):
...
+                with torch.no_grad():
+                    split_linear.weight.copy_(linear.weight[i : i + self.split_size, :].detach())
+                self.split_linear_layers.append(split_linear)
+
+        elif serialization_mode is poptorch.MatMulSerializationMode.ReducingDim:
+            if self.in_features % serialization_factor != 0:
+                raise ValueError(f"{linear.in_features=} must be divisible by {serialization_factor=}")
+
+            self.split_size = self.in_features // serialization_factor
+            for i in range(0, self.out_features, self.split_size):
```
katalinic-gc, line 651: `in_features`?

optimum/graphcore/modeling_utils.py
```diff
@@ -611,51 +611,91 @@ class SplitProjection(torch.nn.Module):
...
-                out += h
+        if self.serialization_mode is poptorch.MatMulSerializationMode.OutputChannels:
+            out = []
+            for i, split_linear_layer in enumerate(self.split_linear_layers):
+                out.append(split_linear_layer(x))
+            out = torch.concat(out, -1)
+
+        elif self.serialization_mode is poptorch.MatMulSerializationMode.ReducingDim:
+            out = self.split_linear_layers[0](x[..., 0 : self.split_size])
+            if len(self.split_linear_layers) > 1:
```
katalinic-gc, line 682: nit, this check is unnecessary, if you loop over `self.split_linear_layers[1:]`, and the resulting list is empty, it will have same effect as not executing at all.

optimum/graphcore/modeling_utils.py
kundaMwiza: I have tested the tests in test_examples.py. I will also check with the notebooks

katalinic-gc (approved):
Thanks

optimum/graphcore/modeling_utils.py
kundaMwiza: I don't think there are any configs that actually use this layer to split the projection layer. Im aware that the whisper finetuning example uses this layer to set which IPU the projection layer will be placed on but since the number of splits is 1 the forward pass will be equivalent to the just using `torch.nn.Linear`.  I think it is safe to land the PR

PR #441, opened 2023-07-04
Integrate whisper large into the existing notebook

alex-coniasse:
Integrate large into the existing Whisper notebook demo instead of making a new one.
(as agreed during the content review meeting)
- add a easy user switch for different sizes (tiny, small, base , large)
Also remove SDK checking for 3.3


notebooks/whisper-example.ipynb
payoto, line 1: Looks good, a few comments


 - Should be an error not a warning if you are going to try and run on a machine with too few IPUs, fail loud and early
 - I would make it auto-select the biggest model based on number of IPUs then make that choice very obvious by printing it in the cell
 - It makes more sense to define the configs first, then choose one of them
 - Once a config is chosen show the user and maybe add a call to action to try out other checkpoints? (This is a point where the user can bring their own weights)
 - I'd consider renaming `model_spec` to `whisper_checkpoint` or `model_checkpoint` (no abbreviations of checkpoint please)
 - Maybe keep the SDK flag for now, it makes the transition easier for us

Can you make your executable cache folder compatible with Paperspace by following these instructions: https://graphcore.atlassian.net/wiki/spaces/PM/pages/3098345498/Writing+a+Paperspace+notebook

alex-coniasse: Thanks Alex! I agree with these points, just a few questions:
- Regarding the warning, I was just worried that would block usage outside of Paperspace environment (when `NUM_AVAILABLE_IPU` is not defined). But I think it doesn't concern a lot of people + it's easy to remove manually.
- For coherence I'm not 100 % sure how to council auto-select with `manual selection` in the last part, when the user is invited to play with the other available models size. 
     Maybe:      `size = auto_select()`
and then we suggest to replace auto_select() manually if you want to test the other models?

notebooks/whisper-example.ipynb
```diff
@@ -121,44 +128,138 @@
...
     "\n",
+    "large_ipu_config = IPUConfig(executable_cache_dir=cache_dir,\n",
+    "                             ipus_per_replica=8,\n",
+    "                             layers_per_ipu=[8, 8, 8, 8, 6, 9, 9, 8])\n",
+    "\n",
+    "configs = {\n",
+    "    \"tiny\": (\"openai/whisper-tiny.en\", \n",
+    "        default_ipu_config),\n",
+    "\n",
+    "    \"small\": (\"openai/whisper-small.en\", \n",
```
katalinic-gc, line 179: Can you swap this with base? The order is tiny, base, small, medium, large.

notebooks/whisper-example.ipynb
```diff
@@ -121,44 +128,138 @@
...
+    "    \"large\": (\"openai/whisper-large-v2\", \n",
+    "        large_ipu_config),\n",
+    "}\n",
+    "\n",
+    "\n",
+    "def select_whisper_config(size: str, custom_checkpoint: str):\n",
+    "    auto_sizes = {4: \"tiny\", 16: \"large\"}\n",
+    "    if size == \"auto\":\n",
+    "        selected_size = auto_sizes[num_available_ipus]\n",
+    "    elif size in [\"tiny\", \"small\", \"base\", \"large\"]:\n",
```
katalinic-gc, line 194: Ditto sorting above

katalinic-gc:
Can you add medium as well? IIRC it will fit on 4 IPUs directly, with `inference_layers_per_ipu=[12, 12, 13, 11]),` giving a more even split.

katalinic-gc (approved):

PR #443, opened 2023-07-05
3.3 Examples CI Fixes

jimypbr:
# What does this PR do?

- Sets the loss scaling for the Glue task which makes training more stable for deberta.
- Enables preprocessing workers in all cases where possible to speed up the data preprocessing.

Fixes # (issue)

## Before submitting


katalinic-gc (approved):
Hopefully there won't be issues with that many workers when running in parallel

jimypbr:
I ran it on a local machine and there won't any issues with the workers.

PR #477, opened 2023-11-25
Fix test assertion in test_trainer.py

slowder:
Fix: replace incorrect call to assertTrue with assertEqual

# What does this PR do?
As originally written, the call to `assertTrue` simply asserts that the `len(preds)` is a truthy value rather than comparing to the second argument.

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).


jimypbr (approved):
LGTM thanks!

